Issue #48, opened 2020-01-11
Logging using spdlog


MacroQuestGitHubMigrationBot:
In GitLab by @brainiac on Jan 10, 2020, 22:52

I already added spdlog to the repo.

Need to do the following:
* setup and initialize spdlog on startup
* replace the backend of DebugSpewXXX functions to use spdlog, and 
* think about how to make this available to plugins (there is a doc on the spdlog wiki about using spdlog across dlls)

PR #553, opened 2022-04-16
Convert DebugSpew to use SPDLOG_DEBUG

Cezero:
Convert DebugSpew to use SPDLOG_DEBUG instead of it's own LogToFile routine.  This eliminates the monolithic debug log and instead puts debug messages in the session specific MacroQuest log, properly timestamped and formatted.

### DIFF
--- a/src/main/MQ2Utilities.cpp
+++ b/src/main/MQ2Utilities.cpp
@@ -35,30 +35,6 @@ namespace mq {
 // Function:    DebugSpew
 // Description: Outputs text to debugger, usage is same as printf ;)
 //***************************************************************************
-
-static void LogToFile(const char* szOutput)
-{
-	FILE* fOut = nullptr;
-
-	const std::filesystem::path pathDebugSpew = std::filesystem::path(mq::internal_paths::Logs) / "DebugSpew.log";
-	const errno_t err = fopen_s(&fOut, pathDebugSpew.string().c_str(), "at");
-
-	if (err || !fOut)
-		return;
-
-#ifdef DBG_CHARNAME
-	char Name[256] = "Unknown";
-	if (pLocalPC)
-	{
-		strcpy_s(Name, pLocalPC->Name);
-	}
-	fprintf(fOut, "%s - ", Name);
-#endif
-
-	fprintf(fOut, "%s\r\n", szOutput);
-	fclose(fOut);
-}
-
 static void DebugSpewImpl(bool always, bool logToFile, const char* szFormat, va_list vaList)
 {
 	if (!always && gFilterDebug)
@@ -73,12 +49,11 @@ static void DebugSpewImpl(bool always, bool logToFile, const char* szFormat, va_
 
 	vsprintf_s(szOutput, theLen, szFormat, vaList);
 
-	strcat_s(szOutput, theLen, "\n");
 	OutputDebugString(szOutput);
 
 	if (logToFile)
 	{
-		LogToFile(szOutput);
+		SPDLOG_DEBUG(szOutput);
 	}
 }
 


src/main/MQ2Utilities.cpp
```diff
@@ -73,12 +49,11 @@ static void DebugSpewImpl(bool always, bool logToFile, const char* szFormat, va_
 
 	vsprintf_s(szOutput, theLen, szFormat, vaList);
 
-	strcat_s(szOutput, theLen, "\n");
 	OutputDebugString(szOutput);
 
 	if (logToFile)
 	{
-		LogToFile(szOutput);
+		SPDLOG_DEBUG(szOutput);
```
brainiac, line 56: What happens if your message has a `{}` in it?

Cezero: It logs as expected:

> D 2022-04-17 13:15:12.693835 [mq] Lua Initializing version 0.100000
> D 2022-04-17 13:15:12.716880 [mq] MQ2EQBC Module Loaded
> D 2022-04-17 13:15:12.716979 [mq] MQ2EQBC::Initializing version {20.041599}
> D 2022-04-17 13:15:12.739378 [mq] MQ2AdvPath Module Loaded
> D 2022-04-17 13:15:12.739470 [mq] Initializing MQ2AdvPath
> D 2022-04-17 13:15:12.761004 [mq] MQ2Cast Module Loaded

Knightly1: He means the literal `{}` not something wrapped in them.

Cezero: what am I missing?

brainiac: Wasn’t sure how it would react given {} is the format placeholder used by fmt

Cezero: Well, it's using `vsprintf_s()`, not fmt, and that part of the code didn't change.  `{}` aren't used by `vsprintf_s`, instead it uses the `%x` parameter formatting similar to `printf()`

brainiac: >  Under the covers, spdlog::default_logger_raw() does use fmt, but only if args are passed, and this code doesn't pass any args, so the fmt call is skipped.

Ah, this is it, then.

brainiac:
Just wanted to leave a comment here so that it doesn't feel ignored -- this is a change that relates to something I've been wanting to do for a while, I intend to pull the change down and play with it a little bit, but haven't had a chance. Hopefully I'll get to it soon.

src/main/MQ2Utilities.cpp
```diff
@@ -73,12 +49,11 @@ static void DebugSpewImpl(bool always, bool logToFile, const char* szFormat, va_
 
 	vsprintf_s(szOutput, theLen, szFormat, vaList);
 
-	strcat_s(szOutput, theLen, "\n");
```
brainiac, line 76: this newline is required for proper output through OutputDebugString.

brainiac:
spdlog is also configured to output to the debug stream, so it'd end up printing messages twice.

Theres a greater need here to revise how all of this works anyways, as all the "debug spew" commands are poorly used and confusing to understand.

Instead, I think that there just needs to be a single entry point, and then filter based on log level. Additionally, we could change it add the file handler based on if we want to log to file (at a process level). we'd also want these to be configurable without requiring a restart.

src/main/MQ2Utilities.cpp
Cezero: it results in extra blank lines in the DebugLog however.  Where does the output from OutputDebugString() go? is that purely for VS debugging?

brainiac: it goes into the process debug stream, which is viewable through tools like DbgView or an attached debugger. I personally use it extensively, so it cannot be made into garbage.

Knightly1:
This is a broader change that needs to be handled -- see #48 .  I'm going to close out this PR since it's not complete and brainiac's concerns about the debug stream haven't been addressed.

Issue #561, opened 2022-05-02
mapfilter aura and campfire not working. Currently being controlled by incorrect toggle


Aipoc76:
Mapfilter aura off does not toggle auras, they are currently being controlled by Mapfilter object.

Mapfilter campfire does nothing, they are being controlled by Mapfilter NPC

This is reproducible.

brainiac:
> This is reproducible.

please include steps to follow to reproduce.

Aipoc76:
NOTE plugin mq2map needs to be loaded.

1 open map. Ensure show is set to all.
2 look at map if auras are not there, type /mapfilter object.  You should then see auras if the object setting was off to start..
3. type /mapfilter aura.  The map will not change.  Auras will still be shown. and the setting should be off. (if on Type /mapfilter aura again)  Once its off you will notice that the auras are still on the map.

4.  The same process applies to campfires.   Only difference is instead of /mapfilter object, they are being controlled by /mapfilter npc

Ortster:
It looks like this is due to the body type changes that were done. The one Aura I tested with (Aura of the Muse) is showing up as a "Construct" body type now.  Instead of what I am guessing was "Untargetable".

Campfire looks like it keys off being Humanoid type, probably moved to something else as well. Don't have access to that to test.

PR #569, opened 2022-06-05
Campfire / Aura Semi Spawn Type Fix

Ortster:
Somewhat Corrects Issue #561

From the recent patch that changed Body Types around it seems Aura and other Aura like abilities got changed around as well.
I have limited access to a variety of Aura but from what I have access too this is what I have found.

Campfire now seem to be "magical" instead of "humanoid".
Normal Aura now seem to be "construct" instead of "untargetable"
Shaman Idol now seem to be "magical" instead of "construct"

These changes move those three to their correct locations, and add race checks for anything relevant to hopefully narrow results. I left the Aura and other 3 items under "utnargetable" untouched just in case. Just like Rune remains under "construct".

I was unable to test the following:
Banner - Moved from "humanoid" to "magical" like campfire was?
Rune - Still "construct" or now "magical" like Idol?
Circle_of / Guardian_Circle / Earthen_Strength - Still "untargetable" or "magical" / "construct" now?

### DIFF
--- a/src/main/MQ2Utilities.cpp
+++ b/src/main/MQ2Utilities.cpp
@@ -6531,8 +6531,8 @@ eSpawnType GetSpawnType(SPAWNINFO* pSpawn)
 			//    return TRIGGER;
 
 		case 100: // Untargetable
-			return UNTARGETABLE
-				;
+			return UNTARGETABLE;
+
 		case 101: // Property Trap
 			return TRAP;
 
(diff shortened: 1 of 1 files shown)


brainiac:
@Ortster i merged this slightly too early - looks like i need the constants that you added to eqlib

Ortster:
@brainiac Was discussing change with @Knightly1, he should be pulling in those to eqlib for me.

brainiac:
I got them. It’s all good.

PR #582, opened 2022-07-26
Console Update

Atomfusion1:
Console Update
This is not to be Pulled But i would like Review so i know its acceptable 

Thanks

### DIFF
--- a/src/main/MQ2ImGuiConsole.cpp
+++ b/src/main/MQ2ImGuiConsole.cpp
@@ -62,9 +62,14 @@ static const int s_userColorFactionLink = USERCOLOR_FACTION_LINK;
 static bool s_dockspaceVisible = true;
 static bool s_consoleVisible = false;
 static bool s_consoleVisibleOnStartup = false;
+static bool s_alphaBool = false;
+static bool s_largeFont = false;
+static bool s_noTitleBar = false;
+static bool s_lockConsole = false;
 static bool s_resetConsolePosition = false;
 static bool s_setFocus = false;
 
+static int s_mouseHoveringConsoleTimer = 0;
 class ImGuiConsole;
 ImGuiConsole* gImGuiConsole = nullptr;
 
@@ -909,6 +914,7 @@ public:
 	std::vector<std::string> m_history;
 	int m_historyPos = -1;    // -1: new line, 0..History.Size-1 browsing history.
 	bool m_scrollToBottom = true;
+	bool AlphaBool = false;
 	std::unique_ptr<ImGuiZepConsole> m_zepEditor;
 
 	ImGuiConsole()
@@ -949,10 +955,15 @@ public:
 
 	void Draw(bool* pOpen)
 	{
-		ImGuiWindowFlags windowFlags = ImGuiWindowFlags_MenuBar;
+		ImGuiWindowFlags windowFlags = 0;
+		if (!s_noTitleBar || s_mouseHoveringConsoleTimer != 0) windowFlags |= ImGuiWindowFlags_MenuBar;
+		if (s_noTitleBar)  windowFlags |= ImGuiWindowFlags_NoTitleBar;
+		if (s_lockConsole) windowFlags |= ImGuiWindowFlags_NoMove;
+		if (s_lockConsole) windowFlags |= ImGuiWindowFlags_NoResize;
 
 		ImGui::SetNextWindowSize(ImVec2(640, 240), ImGuiCond_FirstUseEver);
 		ImGui::PushStyleVar(ImGuiStyleVar_WindowPadding, ImVec2(1, 0));
+		AlphaSetting(AlphaBool);
 
 		if (!ImGui::Begin("MacroQuest Console", pOpen, windowFlags))
 		{
@@ -961,6 +972,7 @@ public:
 			ImGui::PopStyleVar();
 			return;
 		}
+		AlphaBool = MouseOverUpdate();
 
 		// Need to unpop this for the menu.
 		ImGui::PopStyleVar();
@@ -970,8 +982,11 @@ public:
 		// Here we create a context menu only available from the title bar.
 		if (ImGui::BeginMenuBar())
 		{
+			// ImGui::SetWindowFontScale(1.2);
+			AlphaBool = MouseOverUpdate();
 			if (ImGui::BeginMenu("Options"))
 			{
+				AlphaBool = MouseOverUpdate();
 				bool autoScroll = m_zepEditor->GetAutoScroll();
 				if (ImGui::MenuItem("Auto-scroll", nullptr, &autoScroll))
 					m_zepEditor->SetAutoScroll(autoScroll);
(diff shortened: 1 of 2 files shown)


src/main/MQ2ImGuiConsole.cpp
```diff
@@ -909,6 +914,7 @@ class ImGuiConsole
 	std::vector<std::string> m_history;
 	int m_historyPos = -1;    // -1: new line, 0..History.Size-1 browsing history.
 	bool m_scrollToBottom = true;
+	bool AlphaBool = false;
```
brainiac, line 917: as a member variable this should follow the convention and start with `m_`. It should also have a more accurate and descriptive name. I'm not sure what an "AlphaBool" does by reading it.

src/main/MQ2ImGuiTools.h
```diff
@@ -30,6 +30,8 @@ void ImGuiTools_DrawWindowsMenu();
 void InitializeImGuiConsole();
 void ShutdownImGuiConsole();
 void UpdateImGuiConsole();
+bool MouseOverUpdate();
+void AlphaSetting(bool);
```
brainiac, line 34: Only public functions should go into this header. These are functions that are internal to the console implementation.

src/main/MQ2ImGuiConsole.cpp
```diff
@@ -1484,6 +1545,25 @@ static void ConsoleSettings()
 	ImGui::SameLine();
 	mq::imgui::HelpMarker("This feature allows you to automatically show the MacroQuest Console upon load.");
 
+	if (ImGui::Checkbox("No Titlebar", &s_noTitleBar))
+	{
+		WritePrivateProfileBool("MacroQuest", "NoTitlebar", s_noTitleBar, mq::internal_paths::MQini);
+	}
+
+	if (ImGui::Checkbox("Turn Alpha On", &s_alphaBool))
```
brainiac, line 1555: This option needs some re-wording and/or clarification as to what it does. Similarly with the option names. "AlphaOn" doesn't really describe meaningfully what it does if it is on or off.

brainiac: This extends to variable names like `s_alphaBool`

src/main/MQ2ImGuiConsole.cpp
```diff
@@ -1328,6 +1354,41 @@ static void MakeColorGradient(float frequency1, float frequency2, float frequenc
...
+		// This may not be the MQ way of doing it Its just a timer for Mouse over window and drop down but makes it very slick giving a delay to fade
+		// This Actually Covers up another issue with Button Presses as with a button down you are not hoveringoverwindow
+		s_mouseHoveringConsoleTimer = 1200;
+		return true;
+	}
+	else
+	{
+		if (s_mouseHoveringConsoleTimer > 0)
+		{
+			s_mouseHoveringConsoleTimer = s_mouseHoveringConsoleTimer - 1;
```
brainiac, line 1385: This logic is frame rate dependent, and will vary heavily based on how often this function updates.

This needs to be changes to be framerate independent. Look at using std::chrono::stready_clock and measure the elapsed time instead of decrementing a counter.

src/main/MQ2ImGuiConsole.cpp
```diff
@@ -1328,6 +1354,41 @@ static void MakeColorGradient(float frequency1, float frequency2, float frequenc
...
+		ImGui::SetNextWindowBgAlpha(.1f);
+	}
+}
+
+bool MouseOverUpdate()
+{
+	if (ImGui::IsWindowHovered())
+	{
+		// This may not be the MQ way of doing it Its just a timer for Mouse over window and drop down but makes it very slick giving a delay to fade
+		// This Actually Covers up another issue with Button Presses as with a button down you are not hoveringoverwindow
```
brainiac, line 1377: I couldn't understand at all what this comment is trying to say

src/main/MQ2ImGuiConsole.cpp
```diff
@@ -1328,6 +1348,49 @@ static void MakeColorGradient(float frequency1, float frequency2, float frequenc
...
+ImGuiWindowFlags ConsoleFlagSetup()
+{
+	ImGuiWindowFlags windowFlags = 0;
+	if (!s_noTitleBar || s_mouseHoveringConsoleTimer != 0) windowFlags |= ImGuiWindowFlags_MenuBar;
+	if (s_noTitleBar)  windowFlags |= ImGuiWindowFlags_NoTitleBar;
+	if (s_lockConsole) windowFlags |= ImGuiWindowFlags_NoMove | ImGuiWindowFlags_NoResize;
+	return windowFlags;
+}
+
+void AlphaSetting(bool AlphaBool)
```
brainiac, line 1360: Similar feedback on function and variable names, it needs some revising.

However, the whole function can be elimininated and just use a one liner at the call site:

```
ImGui::SetNextWindowBgAlpha(m_alphaBool || !s_alphaBool || s_mouseHoveringConsoleTimer != 0 ? 1.0f : .1f);
```

src/main/MQ2ImGuiConsole.cpp
```diff
@@ -1328,6 +1348,49 @@ static void MakeColorGradient(float frequency1, float frequency2, float frequenc
...
 
 //----------------------------------------------------------------------------
+ImGuiWindowFlags ConsoleFlagSetup()
+{
+	ImGuiWindowFlags windowFlags = 0;
+	if (!s_noTitleBar || s_mouseHoveringConsoleTimer != 0) windowFlags |= ImGuiWindowFlags_MenuBar;
+	if (s_noTitleBar)  windowFlags |= ImGuiWindowFlags_NoTitleBar;
+	if (s_lockConsole) windowFlags |= ImGuiWindowFlags_NoMove | ImGuiWindowFlags_NoResize;
+	return windowFlags;
+}
```
brainiac, line 1358: This function should be member function of the console. Its name should be more indicative of its behavior, so something like `GetConsoleFlags` would be more appropriate.

src/main/MQ2ImGuiConsole.cpp
```diff
@@ -1328,6 +1348,49 @@ static void MakeColorGradient(float frequency1, float frequency2, float frequenc
...
+	{
+		ImGui::SetNextWindowBgAlpha(1.0f);
+	}
+	else 
+	{
+		ImGui::SetNextWindowBgAlpha(.1f);
+	}
+}
+
+bool MouseOverUpdate()
```
brainiac, line 1372: This should be a member funciton too, and have a name that reflects what it does more accurately. suggestion: UpdateMouseHoverConsoleTimer.

src/main/MQ2ImGuiConsole.cpp
```diff
@@ -1020,8 +1029,19 @@ class ImGuiConsole
...
 							m_zepEditor->DoAchievementLinkTest();
 						}
+						if (ImGui::MenuItem("Alpha Control"))
+						{
+							int textFlags = ImGuiInputTextFlags_EnterReturnsTrue
+								| ImGuiInputTextFlags_CallbackCompletion
+								| ImGuiInputTextFlags_CallbackHistory;
+							bool bTextEdit = ImGui::InputText("##Input", m_inputBuffer, IM_ARRAYSIZE(m_inputBuffer), textFlags,
+								[](ImGuiInputTextCallbackData* data)
+								{ return static_cast<ImGuiConsole*>(data->UserData)->TextEditCallback(data); }, this);
```
brainiac, line 1039: I'm not sure whats going on here

brainiac:
not sure why you closed it but i provided some feedback anyways.

PR #590, opened 2022-07-29
ConsoleUpdateV1.2

Atomfusion1:
This Update is for the Console window, Adding additional functionality 
1. Added Settings under Windows-> Console
      A. Lock Window 
      B. Hide TitleBar 
      C. Window Alpha Normal 
      D. Window Alpha Fade 
      E. Console Font Size Adjustment 
      F. UI Selector under console (Can Select Fonts for all Windows and Console text 

Files Changed ImGuiUtils: 
1. Added 2 Additional Font Sizes of Lucon 15 and 17 

MQ2ImGuiConsole:
GetConsolesFlags() Function to handle Flags 
MouseOverUpdate() Function to handle Window Fade Delay from Normal to Fade Condition 
Set m_zepEditor to use Global Font with Selectable Size 
FontNumberFromString() Get Front[#] from Debug String file name 
Created ImGui Items for Console Window, 2 check boxes 3 sliders 1 font control 
Save All Changes to MacroQuest.ini 
Changed Save Location from [Macroquest] to [MQConsole] As now there are 7+ options 
Saving: ConsoleFontOnStartup, ShowMacroQuestConsole, LockConsoleWindow, NoTitlebar, ConsoleFontSize, AlphaFadeOnStartup, AlphaNormalOnStartup
Autoload all options on Startup 
Set Default Values for All Options if Button Pressed

### DIFF
--- a/src/main/MQ2ImGuiConsole.cpp
+++ b/src/main/MQ2ImGuiConsole.cpp
@@ -27,1504 +27,1638 @@
 
 namespace mq {
 
-//----------------------------------------------------------------------------
+	//----------------------------------------------------------------------------
 
-// Indicates that there has been a request to toggle the console.
-bool gbToggleConsoleRequested = false;
+	// Indicates that there has been a request to toggle the console.
+	bool gbToggleConsoleRequested = false;
 
-// Indicates that there has been a request to hide/show the console
-std::optional<bool> gbSetConsoleVisibilityRequest = std::nullopt;
+	// Indicates that there has been a request to hide/show the console
+	std::optional<bool> gbSetConsoleVisibilityRequest = std::nullopt;
 
-static const ImU32 s_defaultColor = Zep::ZepColor(240, 240, 240, 255);
-static ImGuiID s_dockspaceId = 0;
-static ImGuiID s_dockspaceTopSegmentId = 0;
+	static const ImU32 s_defaultColor = Zep::ZepColor(240, 240, 240, 255);
+	static ImGuiID s_dockspaceId = 0;
+	static ImGuiID s_dockspaceTopSegmentId = 0;
 
-// Some plain default colors
-static const ImU32 s_defaultLinkColor = Zep::ZepColor(0, 128, 255);
-static const ImU32 s_defaultLinkColorHover = Zep::ZepColor(255, 255, 128);
+	// Some plain default colors
+	static const ImU32 s_defaultLinkColor = Zep::ZepColor(0, 128, 255);
+	static const ImU32 s_defaultLinkColorHover = Zep::ZepColor(255, 255, 128);
 
-// Some default color constants to patch eq style links
-static const ImU32 s_linkHoverColorDefault = Zep::ZepColor(0, 0, 255);
-static const ImU32 s_linkHoverColorSpam = Zep::ZepColor(0, 255, 0);
-static const ImU32 s_linkHoverColorPlayer = Zep::ZepColor(138, 163, 255);
-static const ImU32 s_linkColorDefault = Zep::ZepColor(0, 255, 255);
-static const ImU32 s_linkColorSpam = Zep::ZepColor(128, 128, 0);
-static const ImU32 s_linkColorPlayer = Zep::ZepColor(0, 0, 0, 0); // use current color
+	// Some default color constants to patch eq style links
+	static const ImU32 s_linkHoverColorDefault = Zep::ZepColor(0, 0, 255);
+	static const ImU32 s_linkHoverColorSpam = Zep::ZepColor(0, 255, 0);
+	static const ImU32 s_linkHoverColorPlayer = Zep::ZepColor(138, 163, 255);
+	static const ImU32 s_linkColorDefault = Zep::ZepColor(0, 255, 255);
+	static const ImU32 s_linkColorSpam = Zep::ZepColor(128, 128, 0);
+	static const ImU32 s_linkColorPlayer = Zep::ZepColor(0, 0, 0, 0); // use current color
 
-static const int s_userColorItemLink = USERCOLOR_LINK;
+	static const int s_userColorItemLink = USERCOLOR_LINK;
 #if IS_EXPANSION_LEVEL(EXPANSION_LEVEL_ROF + 1)
-static const int s_userColorAchievementLink = USERCOLOR_ACHIEVEMENT;
-static const int s_userColorDialogLink = USERCOLOR_DIALOG_LINK;
-static const int s_userColorCommandLink = USERCOLOR_DIALOG_LINK;
-static const int s_userColorFactionLink = USERCOLOR_FACTION_LINK;
+	static const int s_userColorAchievementLink = USERCOLOR_ACHIEVEMENT;
+	static const int s_userColorDialogLink = USERCOLOR_DIALOG_LINK;
+	static const int s_userColorCommandLink = USERCOLOR_DIALOG_LINK;
+	static const int s_userColorFactionLink = USERCOLOR_FACTION_LINK;
 #endif
 
-static bool s_dockspaceVisible = true;
-static bool s_consoleVisible = false;
-static bool s_consoleVisibleOnStartup = false;
-static bool s_resetConsolePosition = false;
-static bool s_setFocus = false;
+	static bool s_dockspaceVisible = true;
+	static bool s_consoleVisible = false;
+	static bool s_consoleVisibleOnStartup = false;
+	static bool s_resetConsolePosition = false;
+	static bool s_setFocus = false;
+	static bool s_isConsoleLocked = false;
+	static bool s_isTitleBarShown = false;
+	static bool s_isMouseInFadeDelay = false;
+	static int s_alphaConsoleNormal = 100;
+	static int s_alphaConsoleFade = 100;
+	static int s_consoleFontSize = 13;
+	static uint16_t s_mouseHoverTimerSecondsDelay = 12;
+	static std::string s_consoleFontName;
 
-class ImGuiConsole;
-ImGuiConsole* gImGuiConsole = nullptr;
+	class ImGuiConsole;
+	ImGuiConsole* gImGuiConsole = nullptr;
 
-static void Strtrim(char* str)
-{
-	char* str_end = str + strlen(str);
-	while (str_end > str && str_end[-1] == ' ')
-		str_end--;
-	*str_end = 0;
-}
-
-unsigned int str_to_hex(char const* p, char const* e) noexcept
-{
-	unsigned int result = 0;
-	while (p != e)
+	static void Strtrim(char* str)
 	{
-		result *= 16;
-		if ('0' <= *p && *p <= '9') { result += *p - '0'; p++; continue; }
-		if ('A' <= *p && *p <= 'F') { result += *p + 10 - 'A'; p++; continue; }
-		if ('a' <= *p && *p <= 'f') { result += *p + 10 - 'a'; p++; continue; }
-		return -1;
+		char* str_end = str + strlen(str);
+		while (str_end > str && str_end[-1] == ' ')
+			str_end--;
+		*str_end = 0;
 	}
 
-	return result;
-}
-
-ImU32 str_to_col(std::string_view str)
-{
-	if (str.length() != 7 || str[0] != '#') { return 0; }
-
-	auto r = str_to_hex(str.data() + 1, str.data() + 3);
-	auto g = str_to_hex(str.data() + 3, str.data() + 5);
-	auto b = str_to_hex(str.data() + 5, str.data() + 7);
-
-	return IM_COL32(r, g, b, 255);
-}
-
-static void MakeColorGradient(float frequency1, float frequency2, float frequency3,
-	float phase1, float phase2, float phase3,
-	float center = 128, float width = 127, int length = 50);
-
-static std::pair<std::string_view, ImU32> ParseColorTags(std::string_view line, std::vector<ImU32>& colorStack,
-	ImU32 defaultColor)
-{
-	size_t length = line.length();
-	const char* pos = line.data();
-	const char* end = line.data() + length;
-
-	ImU32 color = defaultColor;
-
-	// skip over the \a
-	if (*(pos++) != '\a')
-		return { {}, color };
-
-	if (pos == end) return { {}, color };
-
-	bool dark = false;
-
-	// clear
-	if (*pos == 'x')
+	unsigned int str_to_hex(char const* p, char const* e) noexcept
 	{
-		pos++;
-
-		if (!colorStack.empty())
+		unsigned int result = 0;
+		while (p != e)
 		{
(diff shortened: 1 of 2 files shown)


brainiac (changes requested):
we have a bit of an issue here in that the whole file has been reformatted. Please pay attention to the changes in your commits before you make them. I can't review this in its current state. Please revert the formatting changes.

src/main/MQ2ImGuiConsole.cpp
```diff
@@ -27,1504 +27,1638 @@
...
 		}
 
-		// Need to unpop this for the menu.
-		ImGui::PopStyleVar();
+		void MouseOverUpdate()
+		{
+			if (ImGui::IsWindowHovered())
+			{
+				m_pr_StartTime = std::chrono::steady_clock::now();
+				s_isMouseInFadeDelay = true;
```
brainiac, line 973: why is this a static variable but the one above it is a member variable?

src/main/MQ2ImGuiConsole.cpp
```diff
@@ -27,1504 +27,1638 @@
...
-		if (!ImGui::Begin("MacroQuest Console", pOpen, windowFlags))
+		void AddWriteChatColorLog(const char* line, ImU32 defaultColor = s_defaultColor, bool newline = false)
 		{
-			ImGui::End();
+			m_zepEditor->AppendFormattedText(line, defaultColor, newline);
+		}
 
-			ImGui::PopStyleVar();
-			return;
+		ImGuiWindowFlags GetConsolesFlags()
```
brainiac, line 959: this doesn't really need to be a function, but theres a typo in its name anyhow.

src/main/MQ2ImGuiConsole.cpp
```diff
@@ -27,1504 +27,1638 @@
...
+		void MouseOverUpdate()
+		{
+			if (ImGui::IsWindowHovered())
+			{
+				m_pr_StartTime = std::chrono::steady_clock::now();
+				s_isMouseInFadeDelay = true;
+			}
+			else
+			{
+				if (std::chrono::duration_cast<std::chrono::seconds>(std::chrono::steady_clock::now() - m_pr_StartTime ) < std::chrono::seconds(s_mouseHoverTimerSecondsDelay))
```
brainiac, line 977: s_mouseHoverTimerSecondsDelay should be stored as a std::chrono::seconds, not a uint16_t. then you wouldn't need all these casts.

src/main/MQ2ImGuiConsole.cpp
```diff
@@ -27,1504 +27,1638 @@
...
+				{
+					s_isMouseInFadeDelay = true;
+				}
+				else
+				{
+					s_isMouseInFadeDelay = false;
+				}
```
brainiac, line 984: this is basically `if (a) { return true } else { return false }
just assign the variable to the result of the conditio.

src/main/MQ2ImGuiConsole.cpp
```diff
@@ -27,1504 +27,1638 @@
...
-	if (!_stricmp("hide", szCommand))
+	static void ConsoleSettings()
 	{
-		gbSetConsoleVisibilityRequest = false;
-		return;
-	}
+		ImGuiIO& io = ImGui::GetIO();
+		if (ImGui::Checkbox("Show Console on Load", &s_consoleVisibleOnStartup))
+		{
+			WritePrivateProfileBool("MQConsole", "ShowMacroQuestConsole", s_consoleVisibleOnStartup, mq::internal_paths::MQini);
```
brainiac, line 1534: you shouldn't change the names of existin gsettings.

src/main/MQ2ImGuiConsole.cpp
```diff
@@ -27,1504 +27,1638 @@
...
-	if (!_stricmp("show", szCommand))
+	int FontNumberFromString(std::string fontString)
 	{
-		gbSetConsoleVisibilityRequest = true;
-		return;
+		ImGuiIO& io = ImGui::GetIO();
+		for (int i = 0; i < io.Fonts->Fonts.Size; i++)
+		{
+			std::string fontstr = io.Fonts->Fonts[i]->GetDebugName();
+			if (strcmp(fontstr.c_str(), fontString.c_str()) == 0)
```
brainiac, line 1521: you took a const char*, turned it into a std::string, and then took the const char* back out.

This function can take a string_view and use ci_equals from the mq string library.

function name isn't correct. "FontNumberFromString" - what about the font number? what are we doing with it? Getting it? converting it? etc

src/main/MQ2ImGuiConsole.cpp
```diff
@@ -27,1504 +27,1638 @@
...
-	if (ImGui::Checkbox("Show Console on Load", &s_consoleVisibleOnStartup))
-	{
-		WritePrivateProfileBool("MacroQuest", "ShowMacroQuestConsole", s_consoleVisibleOnStartup, mq::internal_paths::MQini);
-	}
+		if (ImGui::Checkbox("Lock Window", &s_isConsoleLocked))
+		{
+			WritePrivateProfileBool("MQConsole", "LockConsoleWindow", s_isConsoleLocked, mq::internal_paths::MQini);
+		}
+		ImGui::SameLine();
+		mq::imgui::HelpMarker("This feature allows you to automatically lock the MacroQuest Console.");
```
brainiac, line 1545: this doesn't tell me what locking the console means.

src/main/MQ2ImGuiConsole.cpp
```diff
@@ -27,1504 +27,1638 @@
...
 
-	ImGui::SameLine();
-	mq::imgui::HelpMarker("This feature allows you to automatically show the MacroQuest Console upon load.");
+		if (ImGui::Checkbox("Hide Titlebar", &s_isTitleBarShown))
+		{
+			WritePrivateProfileBool("MQConsole", "NoTitlebar", s_isTitleBarShown, mq::internal_paths::MQini);
+		}
+		ImGui::SameLine();
+		mq::imgui::HelpMarker("This feature allows you to hide MacroQuest Menu Bar.");
```
brainiac, line 1552: +the, Menu Bar should be lower cased. Should probably make some mention of the title bar since the option is called "Hide Titlebar"

src/main/MQ2ImGuiConsole.cpp
```diff
@@ -27,1504 +27,1638 @@
...
-	ImGui::NewLine();
+		// Setup Sliders 
+		ImGui::PushStyleVar(ImGuiStyleVar_FramePadding, ImVec2(2.0f, 1.0f));
+		ImGui::PushItemWidth(120);
 
-	if (ImGui::Button("Clear Saved Console Settings"))
-	{
-		s_consoleVisibleOnStartup = false;
-		WritePrivateProfileBool("MacroQuest", "ShowMacroQuestConsole", s_consoleVisibleOnStartup, mq::internal_paths::MQini);
+		ImGui::Text("Alpha Normal");
```
brainiac, line 1558: These options need some explanation as i'm not sure what i'm looking at based on their names.

Also, these sliders have labels on the right side, might want to use the space that is reserved for the labels or else remove the extra spacing.

src/main/MQ2ImGuiConsole.cpp
```diff
@@ -27,1504 +27,1638 @@
...
+			s_alphaConsoleNormal = 100;
+			WritePrivateProfileInt("MQConsole", "AlphaNormalOnStartup", s_alphaConsoleNormal, mq::internal_paths::MQini);
+			s_alphaConsoleFade = 100;
+			WritePrivateProfileInt("MQConsole", "AlphaFadeOnStartup", s_alphaConsoleFade, mq::internal_paths::MQini);
+			s_consoleFontSize = 13;
+			WritePrivateProfileInt("MQConsole", "ConsoleFontSize", s_consoleFontSize, mq::internal_paths::MQini);
+			s_consoleFontName = "lucon.ttf, 13px";
+			s_consoleFontSize = 13;
+			int fontNumber = FontNumberFromString(s_consoleFontName);
+			ImGui::GetIO().FontDefault = ImGui::GetIO().Fonts->Fonts[fontNumber];
```
brainiac, line 1613: the console font is not the default font.

src/main/MQ2ImGuiConsole.cpp
```diff
@@ -27,1504 +27,1638 @@
...
-		// Here we create a context menu only available from the title bar.
-		if (ImGui::BeginMenuBar())
+		void Draw(bool* pOpen)
 		{
-			if (ImGui::BeginMenu("Options"))
+			ImGuiWindowFlags windowFlags = GetConsolesFlags();
+
+			ImGui::SetNextWindowSize(ImVec2(640, 240), ImGuiCond_FirstUseEver);
+			ImGui::PushStyleVar(ImGuiStyleVar_WindowPadding, ImVec2(1, 0));
+			ImGui::SetNextWindowBgAlpha(s_isMouseInFadeDelay ? s_alphaConsoleNormal * .01f : s_alphaConsoleFade * .01f);
```
brainiac, line 994: factor out the math from this condition

src/main/MQ2ImGuiConsole.cpp
```diff
@@ -27,1504 +27,1638 @@
...
 	{
-		fmt::basic_memory_buffer<char> buf;
-		fmt::format_to(fmt::appender(buf), fmt, args...);
+	public:
+		char m_inputBuffer[2048];
+		ImVector<const char*> m_commands;
+		std::vector<std::string> m_history;
+		int m_historyPos = -1;    // -1: new line, 0..History.Size-1 browsing history.
+		bool m_scrollToBottom = true;
+		std::chrono::steady_clock::time_point m_pr_StartTime;
```
brainiac, line 920: what is the "pr" signifying?

src/main/MQ2ImGuiConsole.cpp
```diff
@@ -27,1504 +27,1638 @@
...
+			ImGuiWindowFlags windowFlags = 0;
+			if (!s_isTitleBarShown || s_isMouseInFadeDelay) windowFlags |= ImGuiWindowFlags_MenuBar;
+			if (s_isTitleBarShown)  windowFlags |= ImGuiWindowFlags_NoTitleBar;
+			if (s_isConsoleLocked) windowFlags |= ImGuiWindowFlags_NoMove | ImGuiWindowFlags_NoResize;
+			return windowFlags;
 		}
 
-		// Need to unpop this for the menu.
-		ImGui::PopStyleVar();
+		void MouseOverUpdate()
```
brainiac, line 968: this function does something with side effects, it should be named like a verb.

src/main/MQ2ImGuiConsole.cpp
```diff
@@ -27,1504 +27,1638 @@
...
+			ImGui::PushStyleVar(ImGuiStyleVar_ItemSpacing, ImVec2(0, 0));
 
-		ImGui::PopFont();
-		ImGui::PopStyleColor();
-		ImGui::PopStyleVar(2);
+			ImVec2 contentSize = ImGui::GetContentRegionAvail();
+			contentSize.y -= footer_height_to_reserve;
 
+			// Set Font for ZepEditor
+			m_zepEditor->mq::imgui::ImGuiZepEditor::SetFont(Zep::ZepTextType::Text, ImGui::GetFont(), s_consoleFontSize);
```
brainiac, line 1097: SetFont should only be called if the font is being changed.

`m_zepEditor->SetFont`, the whole function scope doesn't need to be qualified.

src/main/MQ2ImGuiConsole.cpp
```diff
@@ -27,1504 +27,1638 @@
...
+		s_consoleVisibleOnStartup = GetPrivateProfileBool("MQConsole", "ShowMacroQuestConsole", false, mq::internal_paths::MQini);
+		s_consoleVisible = s_consoleVisibleOnStartup;
+		s_alphaConsoleFade = GetPrivateProfileInt("MQConsole", "AlphaFadeOnStartup", 100, mq::internal_paths::MQini);
+		s_alphaConsoleNormal = GetPrivateProfileInt("MQConsole", "AlphaNormalOnStartup", 100, mq::internal_paths::MQini);
+		s_consoleFontSize = GetPrivateProfileInt("MQConsole", "ConsoleFontSize", 13, mq::internal_paths::MQini);
+		s_isTitleBarShown = GetPrivateProfileBool("MQConsole", "NoTitleBar", false, mq::internal_paths::MQini);
+		s_isConsoleLocked = GetPrivateProfileBool("MQConsole", "LockConsoleWindow", false, mq::internal_paths::MQini);
+
+		// load saved font 
+		s_consoleFontName = GetPrivateProfileString("MQConsole", "ConsoleFontOnStartup", "Lucon.ttf, 13px", mq::internal_paths::MQini);
```
brainiac, line 1630: the default here is capitalized, but FontNumberFromString is using a case sensitive comparison.

src/main/MQ2ImGuiConsole.cpp
```diff
@@ -27,1504 +27,1638 @@
...
+			s_isTitleBarShown = false;
+			WritePrivateProfileBool("MQConsole", "NoTitleBar", s_isTitleBarShown, mq::internal_paths::MQini);
+			s_alphaConsoleNormal = 100;
+			WritePrivateProfileInt("MQConsole", "AlphaNormalOnStartup", s_alphaConsoleNormal, mq::internal_paths::MQini);
+			s_alphaConsoleFade = 100;
+			WritePrivateProfileInt("MQConsole", "AlphaFadeOnStartup", s_alphaConsoleFade, mq::internal_paths::MQini);
+			s_consoleFontSize = 13;
+			WritePrivateProfileInt("MQConsole", "ConsoleFontSize", s_consoleFontSize, mq::internal_paths::MQini);
+			s_consoleFontName = "lucon.ttf, 13px";
+			s_consoleFontSize = 13;
```
brainiac, line 1611: this is set twice.

src/main/MQ2ImGuiConsole.cpp
```diff
@@ -27,1504 +27,1638 @@
...
+		ImGui::PopStyleVar();
+
+		// Select UI font for console
+		ImGui::Text("UI Fonts Selector");
+		ImGui::ShowFontSelector("##FontsSelector");
+
+		// Check If UI Font changes and save it
+		std::string fontstr = ImGui::GetFont()->GetDebugName();
+		std::string consoleDebugName = s_consoleFontName;
+		if (fontstr != consoleDebugName)
```
brainiac, line 1586: use ci_equals so you don't have to copy these into std::strings

src/main/MQ2ImGuiConsole.cpp
brainiac: you should detect if the font is not found and fall back to the default console font (mq::imgui::ConsoleFont)

Atomfusion1:
I messed this Pull Request up, All changes have been made and are now in ConsoleUpdateV1.3 Thanks for your Advice

PR #593, opened 2022-07-30
Console v1.3

Atomfusion1:
This Update is for the Console window, Adding additional functionality

Added Settings under Windows-> Console
A. Lock Window
B. Hide TitleBar
C. Window Alpha Normal
D. Window Alpha Fade
E. Console Font Size Adjustment
F. UI Selector under console (Can Select Fonts for all Windows and Console text
Files Changed ImGuiUtils:

Added 2 Additional Font Sizes of Lucon 15 and 17
MQ2ImGuiConsole:
GetConsolesFlags() Function to handle Flags
MouseOverUpdate() Function to handle Window Fade Delay from Normal to Fade Condition
Set m_zepEditor to use Global Font with Selectable Size
FontNumberFromString() Get Front[#] from Debug String file name
Created ImGui Items for Console Window, 2 check boxes 3 sliders 1 font control
Save All Changes to MacroQuest.ini
Changed Save Location from [Macroquest] to [MQConsole] As now there are 7+ options
Saving: ConsoleFontOnStartup, ShowMacroQuestConsole, LockConsoleWindow, NoTitlebar, ConsoleFontSize, AlphaFadeOnStartup, AlphaNormalOnStartup
Autoload all options on Startup
Set Default Values for All Options if Button Pressed

All previous Pull Request have been Changed/Attempted to be fixed or modified. 
There is still one Issue That i am having with setup and use of accessor in MQ2ImGuiConsole.cpp line 879

### DIFF
--- a/src/imgui/ImGuiUtils.cpp
+++ b/src/imgui/ImGuiUtils.cpp
@@ -86,8 +86,6 @@ void ConfigureDefaultFont(ImFontAtlas* atlas)
 			ImFontConfig consoleFontConfig;
 			consoleFontConfig.OversampleH = consoleFontConfig.OversampleV = 3;
 			ConsoleFont = atlas->AddFontFromFileTTF(szConsoleFont, 13.0f, &consoleFontConfig);
-			atlas->AddFontFromFileTTF(szConsoleFont, 15.0f, &consoleFontConfig);
-			atlas->AddFontFromFileTTF(szConsoleFont, 17.0f, &consoleFontConfig);
 		}
 	}
 
--- a/src/imgui/ImGuiUtils.cpp
+++ b/src/imgui/ImGuiUtils.cpp
@@ -86,6 +86,8 @@ void ConfigureDefaultFont(ImFontAtlas* atlas)
 			ImFontConfig consoleFontConfig;
 			consoleFontConfig.OversampleH = consoleFontConfig.OversampleV = 3;
 			ConsoleFont = atlas->AddFontFromFileTTF(szConsoleFont, 13.0f, &consoleFontConfig);
+			atlas->AddFontFromFileTTF(szConsoleFont, 15.0f, &consoleFontConfig);
+			atlas->AddFontFromFileTTF(szConsoleFont, 17.0f, &consoleFontConfig);
 		}
 	}
 
--- a/src/main/MQ2ImGuiConsole.cpp
+++ b/src/main/MQ2ImGuiConsole.cpp
@@ -65,11 +65,6 @@ static bool s_consoleVisibleOnStartup = false;
 static bool s_resetConsolePosition = false;
 static bool s_setFocus = false;
 
-static bool s_isConsoleLocked = false;
-static bool s_isTitleBarShown = false;
-static int s_alphaConsoleNormal = 100;
-static int s_alphaConsoleFade = 100;
-
 class ImGuiConsole;
 ImGuiConsole* gImGuiConsole = nullptr;
 
@@ -916,6 +911,10 @@ public:
 	int m_historyPos = -1;    // -1: new line, 0..History.Size-1 browsing history.
 	bool m_scrollToBottom = true;
 	bool m_isConsoleInFadeDelay;
+	inline static bool m_isTitleBarShown = true;
+	inline static int m_alphaConsoleNormal = 100;
+	inline static int m_alphaConsoleFade = 100;
+	inline static bool m_isConsoleLocked = false;
 
 	std::unique_ptr<ImGuiZepConsole> m_zepEditor;
 
@@ -975,13 +974,13 @@ public:
 	void Draw(bool* pOpen)
 	{
 		ImGuiWindowFlags windowFlags = 0;
-		if (!s_isTitleBarShown || m_isConsoleInFadeDelay) windowFlags |= ImGuiWindowFlags_MenuBar;
-		if (s_isTitleBarShown)  windowFlags |= ImGuiWindowFlags_NoTitleBar;
-		if (s_isConsoleLocked) windowFlags |= ImGuiWindowFlags_NoMove | ImGuiWindowFlags_NoResize;
+		if (!m_isTitleBarShown || m_isConsoleInFadeDelay) windowFlags |= ImGuiWindowFlags_MenuBar;
+		if (m_isTitleBarShown)  windowFlags |= ImGuiWindowFlags_NoTitleBar;
+		if (m_isConsoleLocked) windowFlags |= ImGuiWindowFlags_NoMove | ImGuiWindowFlags_NoResize;
 
 		ImGui::SetNextWindowSize(ImVec2(640, 240), ImGuiCond_FirstUseEver);
 		ImGui::PushStyleVar(ImGuiStyleVar_WindowPadding, ImVec2(1, 0));
-		ImGui::SetNextWindowBgAlpha(.01f * (m_isConsoleInFadeDelay ? s_alphaConsoleNormal : s_alphaConsoleFade));
+		ImGui::SetNextWindowBgAlpha(.01f * (m_isConsoleInFadeDelay ? m_alphaConsoleNormal : m_alphaConsoleFade));
 
 		if (!ImGui::Begin("MacroQuest Console", pOpen, windowFlags))
 		{
@@ -1507,6 +1506,7 @@ void MQConsoleCommand(SPAWNINFO* pChar, char* Line)
 
 static void ConsoleSettings()
 {
+	ImGuiConsole Console;
 	if (ImGui::Checkbox("Show Console on Load", &s_consoleVisibleOnStartup))
 	{
 		WritePrivateProfileBool("MQConsole", "ShowMacroQuestConsole", s_consoleVisibleOnStartup, mq::internal_paths::MQini);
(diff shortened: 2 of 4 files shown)


src/main/MQ2ImGuiConsole.cpp
```diff
@@ -569,6 +579,21 @@ class ZepMode_ImGuiConsole : public Zep::ZepMode_Standard
...
+	using namespace std::chrono;
+	if (m_isMouseHovered)
+	{
+		s_mouseHoverStartTimePoint = steady_clock::now();
+		return true;
+	}
+	else
+	{
+		auto m_mouseOverDuration = duration_cast<seconds>(steady_clock::now() - s_mouseHoverStartTimePoint);
+		return (m_mouseOverDuration < s_mouseHoverTimerSecondsDelay) ? true : false;
```
brainiac, line 593: using ternary for boolean expression is completely unnecessary. The condition itself is a boolean, so just return it.

src/main/MQ2ImGuiConsole.cpp
```diff
@@ -569,6 +579,21 @@ class ZepMode_ImGuiConsole : public Zep::ZepMode_Standard
 	}
 };
 
+bool TimedMouseHoveredEvent(bool m_isMouseHovered)
```
brainiac, line 582: `m_` is the prefix for a member variable, of which this is not.

src/main/MQ2ImGuiConsole.cpp
```diff
@@ -64,6 +64,16 @@ static bool s_consoleVisible = false;
...
+static bool s_isConsoleLocked = false;
+static bool s_isTitleBarShown = false;
+static bool s_isMouseInFadeDelay = false;
+static bool s_updateConsoleFont = false;
+static int s_alphaConsoleNormal = 100;
+static int s_alphaConsoleFade = 100;
+static int s_consoleFontSize = 13;
+static auto s_mouseHoverTimerSecondsDelay = std::chrono::seconds(10);
+static std::chrono::steady_clock::time_point s_mouseHoverStartTimePoint;
+static std::string s_consoleFontName;
```
brainiac, line 76: I suspect a lot of these should be member variables. The reason why some of these exist as statics is because they hold state external to the console that is used to control it.

Most of this seems internal to the console and don't need to be here.

src/imgui/ImGuiUtils.cpp
```diff
@@ -86,6 +86,8 @@ void ConfigureDefaultFont(ImFontAtlas* atlas)
 			ImFontConfig consoleFontConfig;
 			consoleFontConfig.OversampleH = consoleFontConfig.OversampleV = 3;
 			ConsoleFont = atlas->AddFontFromFileTTF(szConsoleFont, 13.0f, &consoleFontConfig);
+			atlas->AddFontFromFileTTF(szConsoleFont, 15.0f, &consoleFontConfig);
+			atlas->AddFontFromFileTTF(szConsoleFont, 17.0f, &consoleFontConfig);
```
brainiac, line 90: creating these and never using them is a significant waste.

src/main/MQ2ImGuiConsole.cpp
```diff
@@ -947,21 +974,31 @@ class ImGuiConsole
...
 	}
 
+	ImGuiWindowFlags GetConsoleFlags()
+	{
+		ImGuiWindowFlags windowFlags = 0;
+		if (!s_isTitleBarShown || s_isMouseInFadeDelay) windowFlags |= ImGuiWindowFlags_MenuBar;
+		if (s_isTitleBarShown)  windowFlags |= ImGuiWindowFlags_NoTitleBar;
+		if (s_isConsoleLocked) windowFlags |= ImGuiWindowFlags_NoMove | ImGuiWindowFlags_NoResize;
+		return windowFlags;
+	}
```
brainiac, line 984: this still doesn't warrant being a function.

src/main/MQ2ImGuiConsole.cpp
```diff
@@ -1054,6 +1092,12 @@ class ImGuiConsole
 		ImVec2 contentSize = ImGui::GetContentRegionAvail();
 		contentSize.y -= footer_height_to_reserve;
 
+		// Set Font for ZepEditor
+		if (s_updateConsoleFont)
+		{
+			m_zepEditor->SetFont(Zep::ZepTextType::Text, ImGui::GetFont(), s_consoleFontSize);
```
brainiac, line 1098: the console specifically uses the console font. This gets the default font. Theres a big mismatch here in the logic.

src/main/MQ2ImGuiConsole.cpp
```diff
@@ -947,21 +974,31 @@ class ImGuiConsole
...
+
 	void Draw(bool* pOpen)
 	{
-		ImGuiWindowFlags windowFlags = ImGuiWindowFlags_MenuBar;
+		ImGuiWindowFlags windowFlags = GetConsoleFlags();
 
 		ImGui::SetNextWindowSize(ImVec2(640, 240), ImGuiCond_FirstUseEver);
 		ImGui::PushStyleVar(ImGuiStyleVar_WindowPadding, ImVec2(1, 0));
-
+		ImGui::SetNextWindowBgAlpha(.01f * (s_isMouseInFadeDelay ? s_alphaConsoleNormal : s_alphaConsoleFade));
```
brainiac, line 993: if this doesn't need fading we probably can skip calling this at all

src/main/MQ2ImGuiConsole.cpp
```diff
@@ -1068,13 +1112,11 @@ class ImGuiConsole
 		ImGui::SetCursorPosY(ImGui::GetCursorPosY() + 4);
 		ImGui::PushItemWidth(ImGui::GetContentRegionAvailWidth());
 		ImGui::PushStyleColor(ImGuiCol_FrameBg, ImVec4(0.00f, 0.00f, 0.00f, 0.00f));
-		ImGui::PushFont(mq::imgui::ConsoleFont);
```
brainiac, line 1071: As mentioned above, you've got a logic error in your understanding of how the fonts are being used. We use a variable-width font for all the UI and a monospace font for the console. With your changes, everything is now using the monospace font which is not ok.

If you want to change fonts, you need to change the console font.

brainiac (changes requested):
There are a LOT of font related issues. I suggest taking out the font changes and focusing on one thing at a time.

brainiac:
For the hover thing that we talked about briefly:

The check for being hovered can go where it needs in Render, but it should assign its value to a member variable of ImGuiZepEditor, say `bool m_hovered`

Then, in the class for ImGuiZepEditor add an accessor to retrieve its value

`bool IsHovered() const { return m_hovered; }`

src/main/MQ2ImGuiConsole.cpp
```diff
@@ -64,6 +64,15 @@ static bool s_consoleVisible = false;
 static bool s_consoleVisibleOnStartup = false;
 static bool s_resetConsolePosition = false;
 static bool s_setFocus = false;
+static bool s_isConsoleLocked = false;
+
+static bool s_isTitleBarShown = false;
```
Atomfusion1, line 69: My Problem Is I cant not figure out how to Move these into Member for instance Into Line 938, ImGuiConsole 
int m_alphaConsoleNormal;
And access it in 1544 	if (ImGui::SliderInt("##Alpha Normal", &m_alphaConsoleNormal, 0, 100))
	{
		WritePrivateProfileInt("MQConsole", "AlphaNormalOnStartup", m_alphaConsoleNormal, mq::internal_paths::MQini);
	}

src/main/MQ2ImGuiConsole.cpp
```diff
@@ -1055,7 +1087,11 @@ class ImGuiConsole
 		contentSize.y -= footer_height_to_reserve;
 
 		m_zepEditor->Render("##ZepConsole", contentSize);
-
+		// Is Mouse over Zep Editor
+		imgui::ImGuiZepEditor obj;
+//		s_isMouseInFadeDelay = TimedMouseHoveredEvent(obj.IsHovered());
+		s_isMouseInFadeDelay = TimedMouseHoveredEvent(ImGui::IsWindowHovered());
+		ImGui::EndChild();
```
Atomfusion1, line 1094: I Keep trying but this Works This way but does not work With the comment line above it, I know this is a scope issue, but i have not been able to access it either Directly or through ImGuiConsole::IsHovered() .. This seems to always return true

Atomfusion1:
Still have Git Wrong

Atomfusion1:
This Update is for the Console window, Adding additional functionality

Added Settings under Windows-> Console
A. Lock Window
B. Hide TitleBar
C. Window Alpha Normal
D. Window Alpha Fade

MQ2ImGuiConsole:
TimedMouseHoveredFader Function to handle Window Fade Delay from Normal to Fade Condition
Created ImGui Items for Console Window, 2 check boxes 2 sliders
Save All Changes to MacroQuest.ini
CheckForPastShowMacroQuestConsole Function to Grab Older MQini and move to new location
Changed Save Location from [Macroquest] to [MQConsole] As now there are 5+ options
Saving: ShowMacroQuestConsole, LockConsoleWindow, NoTitlebar, AlphaFadeOnStartup, AlphaNormalOnStartup
Autoload all options on Startup
Set Default Values for All Options if Button Pressed

Latest Changes:
Removed all Font and Font Size Changes
Code Cleanup
Removed Some Static Variables
SetNextWindowBGAlpha Seems to be Required to be called every frame If i Put a latch on it NoAlpha Console
Found Flag for ImGui::IsWindowHovered(ImGuiHoveredFlags_ChildWindows) Detects mouse in Child Windows

Knightly1:
Closing this one out since the requested changes haven't been addressed.

PR #601, opened 2022-08-07
Store radius values of mapfilters. Fixes #600

### LINKED ISSUE #600: CastRadius/SpellRadius do not save values.
When using "/mapfilter castradius x" or spellradius the circles are correctly drawn on the map but are saved to the ini as =1. Seems to be saved as a bool and not an init. All settings of >0 result in a =1 in the ini. No circles appear on the map upon re-long. Can only assume there being drawn w a radius=1. Manually editing the ini still gets changed back to a =1

aquietone:
- Updates `/mapf` command to output correct radius value instead of just 0/1 for radius filters.
- Updates `/mapf xyzradius #` to store the `#` in the INI instead of just `0/1`.
- Updates plugin init to read the radius value from the INI instead of skipping it. The existing comment here about why these radius weren't being set on load kinda makes sense though...

Not sure if anything extra is needed for target or spell radius, just updated for camp and pull.

### DIFF
--- a/src/plugins/map/MQ2Map.cpp
+++ b/src/plugins/map/MQ2Map.cpp
@@ -263,15 +263,27 @@ PLUGIN_API void InitializePlugin()
 	{
 		MapFilterOption& option = MapFilterOptions[i];
 
-		// If it's the CampRadius or PullRadius, let's not get the last on/off state, lets assume it's off so we don't draw a circle at 0, 0.
-		if (_stricmp(option.szName, "CampRadius") && _stricmp(option.szName, "PullRadius"))
+		option.Enabled = GetPrivateProfileInt("Map Filters", option.szName, option.Default, INIFileName);
+		// If it's the CampRadius or PullRadius, set the radius value as well as the loc
+		if (!_stricmp(option.szName, "CampRadius"))
 		{
-			option.Enabled = GetPrivateProfileInt("Map Filters", option.szName, option.Default, INIFileName);
+			option.Radius = GetPrivateProfileFloat("Map Filters", option.szName, option.Default, INIFileName);
+			if (pLocalPlayer && option.Radius > 0.0f)
+			{
+				CampX = pLocalPlayer->X;
+				CampY = pLocalPlayer->Y;
+			}
 		}
-		else
+		if (!_stricmp(option.szName, "PullRadius"))
 		{
-			option.Enabled = 0;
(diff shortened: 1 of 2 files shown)


src/plugins/map/MQ2Map.cpp
```diff
@@ -263,15 +263,27 @@ PLUGIN_API void InitializePlugin()
...
 		{
-			option.Enabled = 0;
+			option.Radius = GetPrivateProfileFloat("Map Filters", option.szName, option.Default, INIFileName);
+			if (pLocalPlayer && option.Radius > 0.0f)
+			{
+				PullX = pLocalPlayer->X;
+				PullY = pLocalPlayer->Y;
+			}
 		}
+		// How about spell radius or target radius?
```
Knightly1, line 286: If you're adding a todo, use TODO --
```
// TODO:  Investigate spell and target radius loading settings and place them here
```
But if you're not sure at all if they're needed I'd just leave it out.

src/plugins/map/MQ2Map.cpp
```diff
@@ -263,15 +263,27 @@ PLUGIN_API void InitializePlugin()
 	{
 		MapFilterOption& option = MapFilterOptions[i];
 
-		// If it's the CampRadius or PullRadius, let's not get the last on/off state, lets assume it's off so we don't draw a circle at 0, 0.
-		if (_stricmp(option.szName, "CampRadius") && _stricmp(option.szName, "PullRadius"))
+		option.Enabled = GetPrivateProfileInt("Map Filters", option.szName, option.Default, INIFileName);
```
Knightly1, line 266: I know this code was just moved, but it looks like a boolean (GetPrivateProfileBool vs Int)

src/plugins/map/MQ2MapCommands.cpp
```diff
@@ -142,31 +146,69 @@ void MapFilterSetting(SPAWNINFO* pChar, MapFilter nMapFilter, const char* szValu
...
 		}
+	}
+
+	if (szValue)
+	{
+		WritePrivateProfileBool("Map Filters", pMapFilter->szName, pMapFilter->Enabled, INIFileName);
+	}
+}
+
+void MapFilterColorSetting(MapFilter nMapFilter, const char* szValue /* = nullptr */)
```
Knightly1, line 157: I don't think the comment here is necessary or really relevant.

PR #602, opened 2022-08-10
Added a spawns binding for lua

dannuic:


brainiac:
should this be like getAllSpawns and filterSpawns? (two separate functions?) I'm not keen on the non-descript "spawns" as a function name.

src/plugins/lua/bindings/lua_MQBindings.cpp
```diff
@@ -600,6 +646,11 @@ void MQ_RegisterLua_MQBindings(sol::table& mq)
 	);
 
 	mq.set_function("FindTextureAnimation",      &FindTextureAnimation);
+
+	//----------------------------------------------------------------------------
+	// Direct Data Bindings
+	mq.set_function("getAllSpawns", &lua_getAllSpawns);
+	mq.set_function("getFilteredSpawns", &lua_getFilteredSpawns);
```
Knightly1, line 653: I agree with the change to the signatures, but I liked the way you had it originally with the elimination of the duplicate code in the functions themselves.

I was thinking something like
```
if (!predicate || predicate_value(lua_spawn))
```
would let you use the same code and then getAllSpawns is just a filtered spawn with a true predicate.

However, not a big deal.

PR #611, opened 2022-08-21
use IsRadius instead of UsesRadius

aquietone:
Fix mapfilters accidentally falling into the radius filter handling


brainiac:
Need to fix line 74 too

PR #613, opened 2022-09-01
Feature Added. IniExtKey.

livermana:
Allows access to ini files with non unique keys.

Example usage
	return # of instances of the "Self Buff" key in the "Buffs" section
	${IniExtKey[<Path>\Character.ini,Buffs,Self Buff]}

	return the first index in the Buffs section of keys named "Self Buff"
	${IniExtKey[<Path>\Character.ini,Buffs,Self Buff,0]}

	return the second index in the Buffs section of keys named "Self Buff"
	${IniExtKey[<Path>\Character.ini,Buffs,Self Buff,1]}

	return the third index in the Buffs section of keys named "Self Buff", with noparse
	${IniExtKey[<Path>\Character.ini,Buffs,Self Buff,2,noparse]}

### DIFF
--- a/src/main/MQ2Data.cpp
+++ b/src/main/MQ2Data.cpp
@@ -146,7 +146,217 @@ bool dataGameTime(const char* szIndex, MQTypeVar& Ret)
 	Ret.Type = pTimeType;
 	return true;
 }
+/// <summary>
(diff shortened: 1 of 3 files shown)


Knightly1:
This is addressed in 49a02809b3e

PR #637, opened 2022-10-04
Event ID Restrictions

dannuic:
Restricted Creation of events so that multiple events can't have the same ID. Also updated the event/bind calls to return booleans to indicate success or failure.


brainiac:
my only nitpick is that the commit message is too long

src/plugins/lua/LuaEvent.cpp
```diff
@@ -51,12 +51,8 @@ void LuaEventProcessor::AddEvent(std::string_view name, std::string_view express
...
 {
 	RemoveEvents({ std::string(name) });
-	auto it = std::find_if(m_eventDefinitions.begin(), m_eventDefinitions.end(),
-		[&name](const std::unique_ptr<LuaEvent>& event) { return event->GetName() == name; });
-	if (it != m_eventDefinitions.end())
-	{
-		m_eventDefinitions.erase(it);
-	}
+	m_eventDefinitions.erase(std::remove_if(m_eventDefinitions.begin(), m_eventDefinitions.end(), 
+		[&name](const std::unique_ptr<LuaEvent>& event) { return event->GetName() == name; }));
```
brainiac, line 55: erase needs 2nd param `m_eventDefinitions.end()` or you're just erasing the first one.

brainiac (changes requested):
fix call to erase

src/plugins/lua/LuaEvent.cpp
dannuic: Yeah, always forget that. Moot now anyway, the better solution is to fix the bug that allowed the same event ID on multiple events (PR updated)

src/plugins/lua/LuaEvent.cpp
```diff
@@ -43,40 +43,56 @@ LuaEventProcessor::~LuaEventProcessor()
...
+bool LuaEventProcessor::AddEvent(std::string_view name, std::string_view expression, const sol::function& function)
 {
+	// the number of events will always be fairly small, and this is a manual operation.
+	// If this is deemed too slow, the event names can be memoized in a set of string_views.
+	auto it = std::find_if(m_eventDefinitions.begin(), m_eventDefinitions.end(),
+		[&name](const std::unique_ptr<LuaEvent>& event) { return event->GetName() == name; });
+
+	if (it != m_eventDefinitions.end())
+	{
+		LuaError("Cannot create event %.*s, it is already defined.", name.length(), name.data());
```
SirCabby, line 55: does this stop the script?

src/plugins/lua/LuaEvent.cpp
```diff
@@ -43,40 +43,56 @@ LuaEventProcessor::~LuaEventProcessor()
...
-void LuaEventProcessor::RemoveEvent(std::string_view name)
+bool LuaEventProcessor::RemoveEvent(std::string_view name)
 {
 	RemoveEvents({ std::string(name) });
 	auto it = std::find_if(m_eventDefinitions.begin(), m_eventDefinitions.end(),
 		[&name](const std::unique_ptr<LuaEvent>& event) { return event->GetName() == name; });
 	if (it != m_eventDefinitions.end())
 	{
 		m_eventDefinitions.erase(it);
+		return true;
```
SirCabby, line 71: is this list we're iterating through only the ones with matching ids?  If not, how do we know it's the last one?

src/plugins/lua/LuaEvent.cpp
dannuic: It does not. It just shows an error (this is the same as bind)

src/plugins/lua/LuaEvent.cpp
dannuic: I'm not sure I understand. It's finding a single entry that is guaranteed to have a unique name because of the main change in this PR.

dannuic:
Restricted Creation of events so that multiple events can't have the same ID. Also updated the event/bind calls to return booleans to indicate success or failure.

src/plugins/lua/LuaEvent.cpp
SirCabby: ah true, this is now assuming unique keys

brainiac:
where did we land on this? Is this a change we still want to make?

dannuic:
The erase call was removed, this should be ready to merge

PR #641, opened 2022-10-05
Added array conversions to lua

dannuic:


src/main/MQ2DataVars.cpp
```diff
@@ -110,7 +110,7 @@ static bool InitVariableValue(MQTypeVar& TypeVar, const DefaultValueType& defaul
 	return false;
 }
 
-static bool InitArrayValue(CDataArray* pArray, const DefaultValueType& defaultValue)
+static bool InitArrayValue(std::shared_ptr<CDataArray> pArray, const DefaultValueType& defaultValue)
```
brainiac, line 113: const ref

src/plugins/lua/bindings/lua_MQBindings.cpp
```diff
@@ -233,13 +233,39 @@ sol::object lua_MQTypeVar::CallVA(sol::this_state L, sol::variadic_args args) co
 	return Call(lua_join(L, ",", args), L);
 }
 
+static sol::table fill_extent(const std::shared_ptr<datatypes::CDataArray>& arr, int extent, int curr_index, sol::state_view s)
+{
+	auto local_size = arr->GetExtents(extent);
```
brainiac, line 238: what datatype is this?

brainiac:
is converting this to a lua table better than just exposing it as a userdata container? Wondering about the cost/benefit

dannuic:
> is converting this to a lua table better than just exposing it as a userdata container? Wondering about the cost/benefit

at some point, the users will need to bring the userdata into lua space as a table anyway, I would think.

brainiac:
maybe, perhaps. its uncertain. i think you could get away with it never being an actual table. my concern is that people like to make a lot of temporary expressions and this could potentially copy into a table repeatedly and then immediately get thrown away.

would need to see use cases to understand how this would be used.

dannuic:
the nominal use case (as I see it) is to get a table return from some datatype that can be looped over, to replace the inconsistent count + access patterns. If there's a possibility that the loop loops across frames, then you want to make sure the indexing isn't incorrect because the underlying array values shifted

brainiac:
> the nominal use case (as I see it) is to get a table return from some datatype that can be looped over, to replace the inconsistent count + access patterns. If there's a possibility that the loop loops across frames, then you want to make sure the indexing isn't incorrect because the underlying array values shifted

that just means we need a frame lock.

I expect people are going to loop over it like this:

mq.TLO.Foo.Bar()[index]

PR #648, opened 2022-10-12
Implement `time` type using chrono and date

brainiac:
Added support for milliseconds, based on work started by @martinlindhe

- Added the following members to the time type: Milliseconds, MillisecondsSinceEpoch and MillisecondsSinceMidnight
- Set night time range to [7pm, 5am)

Times are represented as local time, otherwise we have to account for time zones.

### DIFF
--- a/src/main/MQ2Utilities.cpp
+++ b/src/main/MQ2Utilities.cpp
@@ -1250,7 +1250,7 @@ void GetGameTime(int* Hour, int* Minute, int* Night)
 	if (Minute)
 		*Minute = eqMinute;
 	if (Night)
-		*Night = ((eqHour < 7) || (eqHour > 18));
+		*Night = eqHour >= 19 || eqHour < 5; // [7pm, 5am)
 }
 
 // ***************************************************************************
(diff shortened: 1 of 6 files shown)


brainiac:
I def pushed the wrong changes, hold on let me fix that.

brainiac:
alright fixed. note to self: don't amend a merge commit and then attempt to rebase it.

src/main/MQ2Utilities.cpp
```diff
@@ -1250,7 +1250,7 @@ void GetGameTime(int* Hour, int* Minute, int* Night)
 	if (Minute)
 		*Minute = eqMinute;
 	if (Night)
-		*Night = ((eqHour < 7) || (eqHour > 18));
+		*Night = eqHour >= 19 || eqHour < 5; // [7pm, 5am)
```
dannuic, line 1253: This varies per server at least, doesn't it? Maybe we should have a night/day cycle config in the future.

dannuic (approved):
LGTM

PR #657, opened 2022-11-10
Add map settings to imgui

SirCabby:


src/plugins/map/MQ2Map.cpp
```diff
@@ -42,6 +42,8 @@ bool HighlightPulseIncreasing = true;
 int HighlightPulseIndex = 0;
 int HighlightPulseDiff = HighlightSIDELEN / 10;
 extern MapObject* gpActiveMapObjects;
+MapFilterOption* mapfilterObjectOptions[100] = { nullptr };
+MapFilterOption* mapFilterOptions[100] = { nullptr };
```
brainiac, line 46: use a vector instead of a statically sized array

src/plugins/map/MQ2MapCommands.cpp
```diff
@@ -1174,146 +1181,340 @@ static void DrawMapSetting_SingleFilter(MapFilter filter)
...
+
+	if (ImGui::Checkbox(option->szName, &option->Enabled))
+		changed = true;
+
+	// Draw a tooltip (?) if this option has requirements from another option list
+	MapFilterOption& requirement = GetMapFilterOption(option->RequiresOption);
+	if (!nested && requirement.szName != nullptr && requirement.ThisFilter != MapFilter::All)
+	{
+		ImGui::SameLine();
+		ImGui::TextDisabled("(?)");
```
brainiac, line 1358: use mq::imgui::HelpMarker

src/plugins/map/MQ2MapCommands.cpp
```diff
@@ -1174,146 +1181,340 @@ static void DrawMapSetting_SingleFilter(MapFilter filter)
...
+	if (option->szHelpString)
+	{
+		ImGui::SameLine();
+		ImGui::TextColored(ImVec4(1.0f, 1.0f, 1.0f, 0.5f), " - %s", option->szHelpString);
+	}
+
+	if (changed && option->IsRegenerateOnChange())
+		regenerate = true;
+	if (changed)
+		WritePrivateProfileBool("Map Filters", option->szName, option->Enabled, INIFileName);
```
brainiac, line 1394: This only works for boolean options

src/plugins/map/MapObject.cpp
```diff
@@ -1051,6 +1052,19 @@ void UpdateDefaultMapLocParams()
...
+{
+	gOverrideMapLocParams.lineSize = gDefaultMapLocParams.lineSize;
+	gOverrideMapLocParams.width = gDefaultMapLocParams.width;
+	gOverrideMapLocParams.color.Red = gDefaultMapLocParams.color.Red;
+	gOverrideMapLocParams.color.Green = gDefaultMapLocParams.color.Green;
+	gOverrideMapLocParams.color.Blue = gDefaultMapLocParams.color.Blue;
+	gOverrideMapLocParams.circleRadius = gDefaultMapLocParams.circleRadius;
+	gOverrideMapLocParams.circleColor.Red = gDefaultMapLocParams.circleColor.Red;
+	gOverrideMapLocParams.circleColor.Green = gDefaultMapLocParams.circleColor.Green;
+	gOverrideMapLocParams.circleColor.Blue = gDefaultMapLocParams.circleColor.Blue;
```
brainiac, line 1065: `gOverrideMapLocParams = gDefaultMapLocParams`

src/plugins/map/MapObject.cpp
```diff
@@ -1080,23 +1094,23 @@ MapObjectMapLoc::~MapObjectMapLoc()
...
-	m_lineSize = params.lineSize;
-	m_width = params.width;
-	m_color = params.color;
-	m_circleRadius = params.circleRadius;
-	m_circleColor = params.circleColor;
+	m_mapLocParams.lineSize = params.lineSize;
+	m_mapLocParams.width = params.width;
+	m_mapLocParams.color = params.color;
+	m_mapLocParams.circleRadius = params.circleRadius;
+	m_mapLocParams.circleColor = params.circleColor;
```
brainiac, line 1101: m_mapLocParams = params

src/plugins/map/MapObject.cpp
```diff
@@ -1080,23 +1094,23 @@ MapObjectMapLoc::~MapObjectMapLoc()
...
-	params.lineSize = m_lineSize;
-	params.width = m_width;
-	params.color = m_color;
-	params.circleRadius = m_circleRadius;
-	params.circleColor = m_circleColor;
+	params.lineSize = m_mapLocParams.lineSize;
+	params.width = m_mapLocParams.width;
+	params.color = m_mapLocParams.color;
+	params.circleRadius = m_mapLocParams.circleRadius;
+	params.circleColor = m_mapLocParams.circleColor;
```
brainiac, line 1113: return m_mapLocParams

src/plugins/map/MapObject.cpp
```diff
@@ -1134,33 +1148,44 @@ void MapObjectMapLoc::UpdateMapLoc()
 	RemoveMapLoc();
 
 	MapViewLine* line = nullptr;
+	MQColor* color = new MQColor();
```
brainiac, line 1151: MQColor is a value type. treat it as an int.

src/plugins/map/MapObject.cpp
```diff
@@ -1169,78 +1194,76 @@ void MapObjectMapLoc::UpdateMapLoc()
...
+	m_index = index;
+	UpdateText();
 }
 
 void MapObjectMapLoc::SetLabel(const std::string& labelText)
 {
-	if (mq::test_and_set(m_labelText, labelText))
-	{
-		UpdateText();
-	}
```
brainiac, line 1243: why are the change checks removed?

src/plugins/map/MapObject.cpp
```diff
@@ -1169,78 +1194,76 @@ void MapObjectMapLoc::UpdateMapLoc()
...
+
+	delete color;
 }
 
 void MapObjectMapLoc::SetIndex(int index)
 {
-	if (mq::test_and_set(m_index, index))
-	{
-		UpdateText();
-	}
```
brainiac, line 1235: why are the change checks removed?

src/plugins/map/MapObject.cpp
```diff
@@ -18,9 +18,10 @@ extern MapObject* pLastTarget;
 MapObject* gpActiveMapObjects = nullptr;
 
 MapLocParams gDefaultMapLocParams;
-static std::map<std::string, MapObjectMapLoc*> sLocationsMap;
-static std::vector<MapObjectMapLoc*> sMapLocs;
-
+std::map<std::string, MapObjectMapLoc*> sLocationsMap;
+std::vector<MapObjectMapLoc*> sMapLocs;
+MapLocParams gOverrideMapLocParams;
+bool mapLocListSelected[100] = { false };
```
brainiac, line 24: use a vector

src/plugins/map/MapObject.cpp
```diff
@@ -1306,6 +1329,13 @@ void DeleteMapLoc(MapObjectMapLoc* mapLoc)
 		sMapLocs.end());
 	sLocationsMap.erase(mapLoc->GetTag());
 
+	// Shift items beyond removal loc left to effectively erase the entry
+	auto index = mapLoc->GetIndex() - 1; // Convert from 1-based display index to 0-based array index
+	for (size_t i = index; i < sMapLocs.size(); i++)
+	{
+		mapLocListSelected[i] = mapLocListSelected[i + 1];
+	}
```
brainiac, line 1337: use a vector and you can just use remove/erase

src/plugins/map/MapObject.cpp
SirCabby: won't that make them both point to the same instance instead of altering the values of one instance?

src/plugins/map/MapObject.h
```diff
@@ -49,6 +49,11 @@ class MapObject
 	virtual SPAWNINFO* GetSpawn() const { return nullptr; }
 	virtual GROUNDITEM* GetGroundItem() const { return nullptr; }
 
+	virtual char* GetLabelText() {
```
brainiac, line 52: this must return const char*

src/plugins/map/MQ2MapCommands.cpp
```diff
@@ -666,15 +669,17 @@ void MapSetLocationCmd(SPAWNINFO* pChar, char* szLine)
 
 		MakeMapLoc(params, label, tag, pos, isDefaultLocSettings);
 
+		std::string labelStr;
```
brainiac, line 672: better off using string_view here

src/plugins/map/MapObject.h
```diff
@@ -178,6 +183,17 @@ struct MapLocParams
...
+
+	MapLocParams GetClone()
+	{
+		MapLocParams result;
+		result.lineSize = lineSize;
+		result.width = width;
+		result.color = MQColor(color.Red, color.Green, color.Blue);
+		result.circleRadius = circleRadius;
+		result.circleColor = MQColor(circleColor.Red, circleColor.Green, circleColor.Blue);
+		return result;
```
brainiac, line 195: return *this

src/plugins/map/MapObject.cpp
SirCabby: I always wanted it updated.  There was a bug I was running down at one point where it was failing this check when it shouldn't have.

src/plugins/map/MapObject.h
```diff
@@ -206,28 +223,30 @@ class MapObjectMapLoc : public MapObject
...
-	MQColor               m_color;
 	std::vector<MapViewLine*> m_lines;
-	float                 m_circleRadius;
-	MQColor               m_circleColor;
 	MapCircle             m_circle;
+	MapLocParams		  m_mapLocParams;
 };
 
+extern std::map<std::string, MapObjectMapLoc*> sLocationsMap;
+extern std::vector<MapObjectMapLoc*> sMapLocs;
```
brainiac, line 244: s prefix means static a.k.a. file local. making these extern means they can't possibly be static anymore.

SirCabby: good call, was named before moved here

src/plugins/map/MQ2MapCommands.cpp
```diff
@@ -24,6 +24,14 @@
 
 #include <fmt/format.h>
 
+const int maxOptionsSize = 100;
+std::stack<MapFilterOption*> optionStack;
```
brainiac, line 28: I don't like this as a global, it is only used in AddMapFilterOptionAsImGuiSetting, and its state outside of that function should always be empty. pass it in as a parameter instead.

src/plugins/map/MQ2MapCommands.cpp
```diff
@@ -506,6 +514,10 @@ void MapSetLocationCmd(SPAWNINFO* pChar, char* szLine)
 			strcpy_s(zloc, temp.substr(0, temp.find(delim)).c_str());
 
 			sprintf_s(tag, "%s,%s,%s", yloc, xloc, zloc);
+
+			std::stringstream targetLabel;
+			targetLabel << "Target: [" << pTarget->Name << "] was here";
+			sprintf_s(label, "%s", _strdup(targetLabel.str().c_str()));
```
brainiac, line 520: this is a memory leak.

src/plugins/map/MapObject.cpp
brainiac: This function technically doesn't need to exist because you can copy with `=`

src/plugins/map/MQ2MapCommands.cpp
```diff
@@ -583,30 +595,25 @@ void MapSetLocationCmd(SPAWNINFO* pChar, char* szLine)
 	MapObjectMapLoc* origLoc = GetMapLocByTag(tag);
 	if (origLoc)
 	{
-		params = origLoc->GetParams();
+		params = origLoc->GetParamsClone();
```
brainiac, line 598: You don't need to have a function that returns a copy. you can just use straight value assignment.

src/plugins/map/MQ2MapCommands.cpp
```diff
@@ -583,30 +595,25 @@ void MapSetLocationCmd(SPAWNINFO* pChar, char* szLine)
...
 	}
 	else
 	{
-		params = gDefaultMapLocParams;
+		params = gDefaultMapLocParams.GetClone();
```
brainiac, line 602: You don't need to have a function that returns a copy. you can just use straight value assignment.

src/plugins/map/MQ2MapCommands.cpp
```diff
@@ -666,15 +669,17 @@ void MapSetLocationCmd(SPAWNINFO* pChar, char* szLine)
 
 		MakeMapLoc(params, label, tag, pos, isDefaultLocSettings);
 
+		std::string labelStr;
+		if (label[0] == 0)
+			labelStr = sMapLocs[sMapLocs.size() - 1]->GetLabelText();
```
brainiac, line 674: does this protect against sMapLocs.size() == 0?

src/plugins/map/MQ2MapCommands.cpp
```diff
@@ -1174,146 +1181,340 @@ static void DrawMapSetting_SingleFilter(MapFilter filter)
...
+			allFilterOptions[i] = &MapFilterOptions[i];
+	}
+
+	// Sort copied array alphabetically
+	std::sort(allFilterOptions, allFilterOptions + maxOptionsSize,
+		[](MapFilterOption* a, MapFilterOption* b) -> bool
+		{
+			if (a == nullptr || a->szName == nullptr) return false;
+			if (b == nullptr || b->szName == nullptr) return true;
+			for (int i = 0; ; i++)
```
brainiac, line 1266: we have lots of string utils, use ci_less (mq/include/base/String.h)

src/plugins/map/MQ2MapCommands.cpp
SirCabby: not protected per se, but it should always be safe.  Line 670 has MakeMapLoc which will guarantee a new loc existing in the array (so minimum size of 1).

brainiac:
in hindsight i probably should've grouped those as a single review

src/plugins/map/MapObject.cpp
brainiac: no, they are value types, not references.

src/plugins/map/MapObject.cpp
brainiac: that means something else changed a value and you're relying on SetIndex to update the text.

src/plugins/map/MQ2MapCommands.cpp
SirCabby: how do I access ci_less's internal struct nocase_compare?

dannuic: to what end? You can just use ci_equals, ci_find_substr, ci_starts_with, or ci_ends_with

dannuic: oh, I see, this is a container. use a ci_unordered container, and just sort it.

SirCabby: ah nvm, I see how it's used now.  I wanted basically this: return ci_less()(a->szName, b->szName); and didn't know the syntax expected.

dannuic: I see you use raw arrays a lot, that's the issue here. Prefer stdlib containers so you don't have to reinvent the wheel on all these kinds of things. I think what you want to do is this:
`std::map<std::string_view, MapFilterOption*, ci_less> allFilterOptions(MapFilterOptions.begin(), MapFilterOptions.end());`

that will give you a map sorted by key using ci_less, in way fewer lines.

SirCabby: yes for some reason I felt compelled to not bring in other types when I worked on this ages ago, and instead tried to reuse existing types that were in play for the plugin.  Brainiac suggested vectors earlier in the PR so I've been trying to update everywhere to using these vectors and like you mentioned, it has already condensed considerably.  Your input is very much appreciated, when I get this is a better state after the refactors, please let me know what you think.  It's still fluid as I try to update some really old patterns from the original submission years ago.

dannuic: well, that's actually not right because MapFilterOptions isn't a map, you'd have to declare the map then loop (with a range loop!) MapFilterOptions and insert that way. Or use std::transform.

brainiac: ci_less()(a, b) was the correct answer

brainiac (changes requested):
marking as changes requested because apparently i forgot to do this the first time

src/plugins/map/MapObject.cpp
SirCabby: no, unfortunately the default value is 0, and I was setting it to 0 for the first time, so the text was never updated for the first entry.

src/plugins/map/MQ2Map.h
```diff
@@ -124,7 +127,7 @@ extern char maphideStr[MAX_STRING];
 extern MQSpawnSearch MapFilterCustom;
 extern MQSpawnSearch MapFilterNamed;
 
-extern MapFilterOption MapFilterOptions[];
+extern std::vector<MapFilterOption> MapFilterOptions;
```
brainiac, line 130: This was changed to a vector but the current usage was an unspecified length array with a terminator (invalid entry). I don't see changes to the code that reads it, so maybe this should stay the same as it was?

src/plugins/map/MapObject.h
```diff
@@ -206,34 +211,58 @@ class MapObjectMapLoc : public MapObject
...
+	void UpdateFromParams(const MapLocParams& params);
+	bool IsCreatedFromDefaults() { return m_isCreatedFromDefaultLoc; }
+	void SetCreatedFromDefaults(bool isCreatedFromDefaults)
+	{
+		m_isCreatedFromDefaultLoc = isCreatedFromDefaults;
+		m_MapLoc->SetCreatedFromDefaults(m_isCreatedFromDefaultLoc);
+	}
+
+private:
+	MapLocParams		  m_mapLocParams;
```
brainiac, line 251: align variable names with spaces. I can see things are off because some are using tabs and they don't align the same.

src/plugins/map/MapObject.cpp
```diff
@@ -1080,25 +1099,20 @@ MapObjectMapLoc::~MapObjectMapLoc()
...
+	m_mapLocParams = params;
+	m_MapLoc->UpdateFromParams(params);
 
-	return params;
+	m_MapLoc->Update(true);
+}
+
+MapLocParams* MapObjectMapLoc::GetParams()
+{
+	return &m_mapLocParams;
```
brainiac, line 1115: this should return const-reference instead of by pointer. The only thing that uses it is dereferencing it anyways, so theres no point to returning a mutable pointer here.

Also make the function `const`

src/plugins/map/MapObject.cpp
```diff
@@ -1134,33 +1148,44 @@ void MapObjectMapLoc::UpdateMapLoc()
 	RemoveMapLoc();
 
 	MapViewLine* line = nullptr;
+	MQColor color = MQColor();
+	color.Red = m_mapLocParams.color.Red;
+	color.Green = m_mapLocParams.color.Green;
+	color.Blue = m_mapLocParams.color.Blue;
+	uint32_t colorARGB = color.ToARGB();
```
brainiac, line 1155: you can just do `colorARGB = m_mapLocParams.color.ToARGB()`

src/plugins/map/MapObject.cpp
```diff
@@ -1134,33 +1148,44 @@ void MapObjectMapLoc::UpdateMapLoc()
...
+
+	// Invert the color (highlight it) temporarily if this loc is selected in the imgui
+	if (m_isSelected)
+	{
+		colorARGB = (0xFFFFFF - colorARGB) | 0xFF000000;
```
brainiac, line 1160: might be a good helper function for MQColor (invert)

src/plugins/map/MapObject.cpp
```diff
@@ -1260,67 +1281,99 @@ void MapObjectMapLoc::UpdateText()
...
 
-MapObjectMapLoc* GetMapLocByTag(const std::string& tag)
+MapLocTemplate* GetMapLocByTag(const std::string& tag)
 {
-	auto iter = sLocationsMap.find(tag);
-	if (iter == sLocationsMap.end())
-		return nullptr;
+	for (auto maplocTemplate : gMapLocTemplates)
+	{
+		if (!strcmp(maplocTemplate->GetMapLoc()->GetTag().c_str(), tag.c_str()))
```
brainiac, line 1299: You're comparing to std::strings here:
`if (maplocTemplate->GetMapLoc()->GetTag() == tag)`

src/plugins/map/MapObject.cpp
```diff
@@ -1260,67 +1281,99 @@ void MapObjectMapLoc::UpdateText()
...
-
-	if (index >= sMapLocs.size())
-		return nullptr;
-
-	return sMapLocs[index];
+	return m_MapLoc;
 }
 
-MapObjectMapLoc* GetMapLocByTag(const std::string& tag)
+MapLocTemplate* GetMapLocByTag(const std::string& tag)
```
brainiac, line 1295: With the correction to string comparison, GetMapLocByTag can be changed to take a `string_view` which will make it work with things that aren't std::string.

src/plugins/map/MQ2MapCommands.cpp
```diff
@@ -664,17 +672,20 @@ void MapSetLocationCmd(SPAWNINFO* pChar, char* szLine)
...
 		CVector3 pos{ std::stof(xloc), std::stof(yloc), std::stof(zloc) };
 
-		MakeMapLoc(params, label, tag, pos, isDefaultLocSettings);
+		MapLocTemplate* newMapLoc = new MapLocTemplate(params, label, tag, pos, isDefaultLocSettings);
+
+		gMapLocTemplates.push_back(newMapLoc);
+
+		UpdateMapLocIndexes();
+
+		std::string_view labelStr = gMapLocTemplates[gMapLocTemplates.size() - 1]->GetMapLoc()->GetLabelText();
```
brainiac, line 683: you can use `newMapLoc` instead of `gMapLocTemplates[gMapLocTemplates.size() - 1]`

src/plugins/map/MapObject.cpp
```diff
@@ -1070,7 +1079,17 @@ MapObjectMapLoc::MapObjectMapLoc(const MapLocParams& params, const std::string&
...
+	Update(true);
+}
+
+MapObjectMapLoc::MapObjectMapLoc()
+{
+	m_isCreatedFromDefaultLoc = false;
+	m_initialized = false;
+	m_index = -1;
+	m_tag = "";
+	m_labelText = "";
```
brainiac, line 1092: strings default initialize to empty, don't need this.

brainiac: actually you're already initializing all of these in the header, don't need any of them.

src/plugins/map/MapObject.cpp
```diff
@@ -968,8 +967,10 @@ void MapObjects_Clear()
 {
 	GroundItemMap.clear();
 	SpawnMap.clear();
-	sLocationsMap.clear();
-	sMapLocs.clear();
+	for (auto maplocTemplate : gMapLocTemplates)
+	{
+		maplocTemplate->ClearReferenceToMapLoc();
```
brainiac, line 972: this should be in the destructor of MapLoc instead, then you don't need to call it explicitly.

src/plugins/map/MapObject.cpp
```diff
@@ -1260,67 +1281,99 @@ void MapObjectMapLoc::UpdateText()
...
 
 void DeleteAllMapLocs()
 {
-	sLocationsMap.clear();
-
-	for (MapObjectMapLoc* loc : sMapLocs)
+	for (auto mapLocTemplate : gMapLocTemplates)
 	{
-		delete loc;
+		delete mapLocTemplate->GetMapLoc();
```
brainiac, line 1310: If this were in the destructor of mapLocTemplate, you don't need to call this.

src/plugins/map/MapObject.cpp
```diff
@@ -1260,67 +1281,99 @@ void MapObjectMapLoc::UpdateText()
...
-		sMapLocs.end());
-	sLocationsMap.erase(mapLoc->GetTag());
+	gMapLocTemplates.erase(std::remove_if(gMapLocTemplates.begin(), gMapLocTemplates.end(),
+		[mapLoc](MapLocTemplate* match) -> bool
+		{
+			if (match->GetMapLoc()->GetTag() == mapLoc->GetTag())
+			{
+				delete match->GetMapLoc();
+				match->ClearReferenceToMapLoc();
+				return true;
```
brainiac, line 1334: If MapLoc's destructor called ClearReferenceToMapLoc, and MapLocTemplate's destructor deleted MapLoc, then you don't need these statements. just return true

src/plugins/map/MQ2MapCommands.cpp
```diff
@@ -664,17 +672,20 @@ void MapSetLocationCmd(SPAWNINFO* pChar, char* szLine)
 		// FIXME: exception handling?
 		CVector3 pos{ std::stof(xloc), std::stof(yloc), std::stof(zloc) };
 
-		MakeMapLoc(params, label, tag, pos, isDefaultLocSettings);
+		MapLocTemplate* newMapLoc = new MapLocTemplate(params, label, tag, pos, isDefaultLocSettings);
```
brainiac, line 677: use unique_ptr instead of raw pointer for MapLocTemplate. Update vector to use std::unique_ptr and then you don't need to delete when removing. If you combine this with the other suggestions, then clearing the container or removing an item will do all the work for you of cleaning things up. also makes it virtually impossible to do something wrong.

src/plugins/map/MQ2MapCommands.cpp
```diff
@@ -1364,6 +1580,387 @@ static void DrawMapSettings_Colors()
...
+			gOverrideMapLocParams.circleColor.Blue = static_cast<uint8_t>(radiusColor.Value.z * 255);
+			gOverrideMapLocParams.circleColor.Green = static_cast<uint8_t>(radiusColor.Value.y * 255);
+			gOverrideMapLocParams.circleColor.Red = static_cast<uint8_t>(radiusColor.Value.x * 255);
+			gOverrideMapLocParams.circleColor.Alpha = 255;
+		}
+
+		gDefaultMapLocParams.circleColor.Blue = static_cast<uint8_t>(radiusColor.Value.z * 255);
+		gDefaultMapLocParams.circleColor.Green = static_cast<uint8_t>(radiusColor.Value.y * 255);
+		gDefaultMapLocParams.circleColor.Red = static_cast<uint8_t>(radiusColor.Value.x * 255);
+		gDefaultMapLocParams.circleColor.Alpha = 255;
```
brainiac, line 1638: you're calculating the same color twice, in these sections.

src/plugins/map/MQ2MapCommands.cpp
```diff
@@ -1364,6 +1580,387 @@ static void DrawMapSettings_Colors()
...
+	ImGui::InputInt("Z", &addMapLocZ, 0);
+	ImGui::InputTextWithHint("Map Label", "<optional>", addMapLocLabel, IM_ARRAYSIZE(addMapLocLabel));
+
+	ImGui::Separator();
+
+	ImGui::TextWrapped("Map Loc List:");
+	ImGui::SameLine();
+	if (ImGui::Button("Select All"))
+	{
+		for (size_t i = 0; i < gMapLocTemplates.size(); i++)
```
brainiac, line 1889: can do ranged-base for loops here

src/plugins/map/MQ2MapCommands.cpp
```diff
@@ -1364,6 +1580,387 @@ static void DrawMapSettings_Colors()
...
+	{
+		ImGui::TableSetupColumn("Index", ImGuiTableColumnFlags_WidthFixed, 50);
+		ImGui::TableSetupColumn("Loc (y,x,z)", ImGuiTableColumnFlags_WidthFixed, 130);
+		ImGui::TableSetupColumn("Label", ImGuiTableColumnFlags_WidthFixed, 300);
+		ImGui::TableSetupColumn("Uses Defaults?");
+		ImGui::TableHeadersRow();
+
+		for (size_t i = 0; i < gMapLocTemplates.size(); i++)
+		{
+			MapObjectMapLoc* maploc = gMapLocTemplates[i]->GetMapLoc();
```
brainiac, line 1916: GetMapLoc can return null

src/plugins/map/MQ2MapCommands.cpp
brainiac: MQcolor has a constructor that takes ImColor.

Theres a lot of these, i'm not commenting on each.

src/plugins/map/MQ2MapCommands.cpp
```diff
@@ -1364,6 +1580,387 @@ static void DrawMapSettings_Colors()
...
+{
+	for (size_t i = 0; i < gMapLocTemplates.size(); i++)
+	{
+		if (gMapLocTemplates[i]->GetMapLoc()->m_isSelected)
+		{
+			return true;
+		}
+	}
+
+	return false;
```
brainiac, line 1666:
```c++
return std::any_of(gMapLocTemplates.begin(), gMapLocTemplates.end(), [](MapLocTemplate* p) { return p->GetMapLoc() && p->GetMapLoc()->IsSelected(); })
```

src/plugins/map/MQ2MapCommands.cpp
```diff
@@ -1364,6 +1580,387 @@ static void DrawMapSettings_Colors()
...
+
+static void DeleteSelectedMapLocs()
+{
+	for (size_t i = 0; i < gMapLocTemplates.size(); i++)
+	{
+		if (gMapLocTemplates[i]->GetMapLoc()->m_isSelected)
+		{
+			DeleteMapLoc(gMapLocTemplates[i--]->GetMapLoc());
+		}
+	}
```
brainiac, line 1677:
```
gMapLocTemplates.erase(
    std::remove_if(gMapLocTemplates.begin(), gMapLocTemplates.end(), [](MapLocTemplate* p) { return p->GetMapLoc() && p->GetMapLoc()->IsSelected(); }),
    gMapLocTemplates.end());
```

src/plugins/map/MQ2MapCommands.cpp
```diff
@@ -1364,6 +1580,387 @@ static void DrawMapSettings_Colors()
...
+
+static void ResetSelectedMapLocsToDefault()
+{
+	for (size_t i = 0; i < gMapLocTemplates.size(); i++)
+	{
+		if (gMapLocTemplates[i]->GetMapLoc()->m_isSelected)
+		{
+			MapObjectMapLoc* thisMapLoc = gMapLocTemplates[i]->GetMapLoc();
+			gMapLocTemplates[i]->SetCreatedFromDefaults(true);
+			gMapLocTemplates[i]->UpdateFromParams(gDefaultMapLocParams);
```
brainiac, line 1688: is this right? thisMapLoc isn't being used here

src/plugins/map/MQ2MapCommands.cpp
```diff
@@ -1364,6 +1580,387 @@ static void DrawMapSettings_Colors()
...
+
+static void ApplyOverridesToSelected(MapLocParams params)
+{
+	for (size_t i = 0; i < gMapLocTemplates.size(); i++)
+	{
+		if (gMapLocTemplates[i]->GetMapLoc()->m_isSelected)
+		{
+			MapObjectMapLoc* thisMapLoc = gMapLocTemplates[i]->GetMapLoc();
+			gMapLocTemplates[i]->SetCreatedFromDefaults(false);
+			gMapLocTemplates[i]->UpdateFromParams(params);
```
brainiac, line 1701: ... nor here

src/plugins/map/MQ2MapCommands.cpp
```diff
@@ -1364,6 +1580,387 @@ static void DrawMapSettings_Colors()
...
+			MapObjectMapLoc* thisMapLoc = gMapLocTemplates[i]->GetMapLoc();
+			gMapLocTemplates[i]->SetCreatedFromDefaults(true);
+			gMapLocTemplates[i]->UpdateFromParams(gDefaultMapLocParams);
+		}
+	}
+}
+
+static void ApplyOverridesToSelected(MapLocParams params)
+{
+	for (size_t i = 0; i < gMapLocTemplates.size(); i++)
```
brainiac, line 1695: use ranged-based for loop, you don't need to keep indexing gMapLocTemplates (same with the function above htis one)

src/plugins/map/MQ2MapCommands.cpp
```diff
@@ -1364,6 +1580,387 @@ static void DrawMapSettings_Colors()
...
+	{
+		ImGui::EndDisabled();
+		disable = false;
+	}
+
+	ImVec2 center = ImGui::GetMainViewport()->GetCenter();
+	ImGui::SetNextWindowPos(center, ImGuiCond_Appearing, ImVec2(0.5f, 0.5f));
+	if (ImGui::BeginPopupModal("Delete?", NULL, ImGuiWindowFlags_AlwaysAutoResize))
+	{
+		ImGui::Text("Are you sure you want to delete all map locs?\n\n");
```
brainiac, line 1730: ImGui::NewLine() instead of making a multiline text

src/plugins/map/MQ2MapCommands.cpp
```diff
@@ -1364,6 +1580,387 @@ static void DrawMapSettings_Colors()
...
+		ImGui::SetItemDefaultFocus();
+		ImGui::SameLine();
+		if (ImGui::Button("Cancel", ImVec2(120, 0)))
+		{
+			ImGui::CloseCurrentPopup();
+		}
+		ImGui::EndPopup();
+	}
+
+	if (pTarget.get() == nullptr)
```
brainiac, line 1747: don't need .get()

src/plugins/map/MQ2MapCommands.cpp
```diff
@@ -1364,6 +1580,387 @@ static void DrawMapSettings_Colors()
...
+		targetStream << "target";
+		if (IsMapLocsStyleOverridden())
+		{
+			targetStream << " size " << gOverrideMapLocParams.lineSize
+				<< " width " << gOverrideMapLocParams.width
+				<< " color " << std::to_string(gOverrideMapLocParams.color.Red) << " " << std::to_string(gOverrideMapLocParams.color.Green) << " " << std::to_string(gOverrideMapLocParams.color.Blue)
+				<< " radius " << gOverrideMapLocParams.circleRadius
+				<< " rcolor " << std::to_string(gOverrideMapLocParams.circleColor.Red) << " " << std::to_string(gOverrideMapLocParams.circleColor.Green) << " " << std::to_string(gOverrideMapLocParams.circleColor.Blue);
+		}
+		char* targetStr = _strdup(targetStream.str().c_str());
```
brainiac, line 1764: _strdup here is a memory leak.

src/plugins/map/MQ2MapCommands.cpp
```diff
@@ -1364,6 +1580,387 @@ static void DrawMapSettings_Colors()
...
+		std::stringstream commandStream;
+		commandStream << addMapLocY << " " << addMapLocX << " " << addMapLocZ;
+		if (IsMapLocsStyleOverridden())
+		{
+			commandStream << " size " << gOverrideMapLocParams.lineSize
+				<< " width " << gOverrideMapLocParams.width
+				<< " color " << std::to_string(gOverrideMapLocParams.color.Red) << " " << std::to_string(gOverrideMapLocParams.color.Green) << " " << std::to_string(gOverrideMapLocParams.color.Blue)
+				<< " radius " << gOverrideMapLocParams.circleRadius
+				<< " rcolor " << std::to_string(gOverrideMapLocParams.circleColor.Red) << " " << std::to_string(gOverrideMapLocParams.circleColor.Green) << " " << std::to_string(gOverrideMapLocParams.circleColor.Blue);
+		}
```
brainiac, line 1855: you don't need to to_string everything here.

src/plugins/map/MQ2MapCommands.cpp
```diff
@@ -1364,6 +1580,387 @@ static void DrawMapSettings_Colors()
...
+		ImGui::SameLine();
+		ImColor radiusColor = gOverrideMapLocParams.circleColor.ToImColor();
+		ImGui::SetNextItemWidth(200);
+		if (ImGui::ColorEdit3("MapLoc Radius Override", &radiusColor.Value.x))
+		{
+			gOverrideMapLocParams.circleColor.Blue = static_cast<uint8_t>(radiusColor.Value.z * 255);
+			gOverrideMapLocParams.circleColor.Green = static_cast<uint8_t>(radiusColor.Value.y * 255);
+			gOverrideMapLocParams.circleColor.Red = static_cast<uint8_t>(radiusColor.Value.x * 255);
+			gOverrideMapLocParams.circleColor.Alpha = 255;
```
brainiac, line 1830: MQColor can convert from ImColor, don't need all this.

src/plugins/map/MQ2MapCommands.cpp
```diff
@@ -1364,6 +1580,387 @@ static void DrawMapSettings_Colors()
...
+		ImGui::DragFloat(" Line Width (1-10, default: 10)", &gOverrideMapLocParams.width, 1.0f, 1, 10, "%.0f", ImGuiSliderFlags_None);
+
+		ImColor mapLocColor = gOverrideMapLocParams.color.ToImColor();
+		ImGui::SetNextItemWidth(200);
+		if (ImGui::ColorEdit3("MapLoc Override", &mapLocColor.Value.x))
+		{
+			gOverrideMapLocParams.color.Blue = static_cast<uint8_t>(mapLocColor.Value.z * 255);
+			gOverrideMapLocParams.color.Green = static_cast<uint8_t>(mapLocColor.Value.y * 255);
+			gOverrideMapLocParams.color.Red = static_cast<uint8_t>(mapLocColor.Value.x * 255);
+			gOverrideMapLocParams.color.Alpha = 255;
```
brainiac, line 1807: MQColor can convert from ImColor, don't need all this.

src/plugins/map/MQ2MapCommands.cpp
```diff
@@ -1169,151 +1182,350 @@ char* FormatMarker(const char* szLine, char* szDest, size_t BufferSize)
...
+
+		mq::imgui::HelpMarker(
+			[&]() -> const char*
+			{
+				std::string requireString = "Requires: ";
+				if (requirement.IsObject() != option->IsObject())
+					requireString += requirement.IsObject() ? "Object Filters -> " : "Options -> ";
+
+				requireString += std::string(requirement.szName);
+				char* requireArray = &requireString[0];
```
brainiac, line 1368: This is returning a pointer to a temporary that is out of scope.

std::requireString no longer exists when the lambda function returns, so using the pointer will be undefined behavior.

src/plugins/map/MQ2MapCommands.cpp
```diff
@@ -1169,151 +1182,350 @@ char* FormatMarker(const char* szLine, char* szDest, size_t BufferSize)
...
+		if (option.ThisFilter != MapFilter::Invalid && option.ThisFilter != MapFilter::All)
+			allFilterOptions.push_back(&option);
+	}
+
+	// Sort copied array alphabetically
+	std::sort(allFilterOptions.begin(), allFilterOptions.end(),
+		[](MapFilterOption* a, MapFilterOption* b) -> bool
+		{
+			if (a == nullptr || a->szName == nullptr) return false;
+			if (b == nullptr || b->szName == nullptr) return true;
```
brainiac, line 1260: in what cases can a/b be nullptr?

src/plugins/map/MQ2MapCommands.cpp
```diff
@@ -1364,6 +1580,387 @@ static void DrawMapSettings_Colors()
...
+				<< " width " << gOverrideMapLocParams.width
+				<< " color " << std::to_string(gOverrideMapLocParams.color.Red) << " " << std::to_string(gOverrideMapLocParams.color.Green) << " " << std::to_string(gOverrideMapLocParams.color.Blue)
+				<< " radius " << gOverrideMapLocParams.circleRadius
+				<< " rcolor " << std::to_string(gOverrideMapLocParams.circleColor.Red) << " " << std::to_string(gOverrideMapLocParams.circleColor.Green) << " " << std::to_string(gOverrideMapLocParams.circleColor.Blue);
+		}
+		if (addMapLocLabel[0] != '\0')
+		{
+			commandStream << " label " << addMapLocLabel;
+		}
+		char* commandStr = _strdup(commandStream.str().c_str());
```
brainiac, line 1860: another _strdup

src/plugins/map/MapObject.h
```diff
@@ -49,6 +49,11 @@ class MapObject
 	virtual SPAWNINFO* GetSpawn() const { return nullptr; }
 	virtual GROUNDITEM* GetGroundItem() const { return nullptr; }
 
+	virtual const char* GetLabelText() {
+		if (m_label == nullptr) return "";
+		return m_label->Label;
+	}
```
brainiac, line 55: Why is this virtual?

src/plugins/map/MapObject.h
```diff
@@ -206,34 +211,58 @@ class MapObjectMapLoc : public MapObject
 
 	virtual bool CanDisplayObject() const override { return true; }
 
+	bool				  m_isSelected = false; // exposed directly for imgui reference
```
brainiac, line 214: alignment

src/plugins/map/MapObject.h
```diff
@@ -206,34 +211,58 @@ class MapObjectMapLoc : public MapObject
...
+		m_MapLoc->SetCreatedFromDefaults(m_isCreatedFromDefaultLoc);
+	}
+
+private:
+	MapLocParams		  m_mapLocParams;
+	std::string           m_label;
+	std::string           m_tag;
+	CVector3              m_pos;
+	bool                  m_isCreatedFromDefaultLoc = false;
+	MapObjectMapLoc*      m_MapLoc = nullptr;
```
brainiac, line 256: nitpick: should be named `m_mapLoc`

src/imgui/ImGuiUtils.cpp
```diff
@@ -210,6 +211,32 @@ void HelpMarker(const char* desc, float width, ImFont* tooltipFont)
...
+	{
+		ImGui::BeginTooltip();
+		ImGui::PushTextWrapPos(width);
+
+		if (tooltipFont)
+		{
+			ImGui::PushFont(tooltipFont);
+		}
+
+		ImGui::TextUnformatted(getText());
```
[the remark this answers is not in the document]
SirCabby: Some follow up questions.  Whenever I leverage the second line, I get a wildly flashing tooltip popup on hover.  I attempted to leverage both string and string_view as return types and that didn't seem to affect this behavior.  Were you suggesting to use string as part of the follow-up instead of string_view?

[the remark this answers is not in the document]
brainiac: > Were you suggesting to use string as part of the follow-up instead of string_view?

Yes.

src/imgui/ImGuiUtils.h
```diff
@@ -43,6 +45,7 @@ IMGUI_API void DrawSplitter(bool split_vertically, float thickness, float* size0
 // Draws a (?) marker with the provided text as its tooltip. Width defines the size of
 // the tooltip box. The height will grow to fit.
 IMGUI_API void HelpMarker(const char* text, float width = 450.0f, ImFont* tooltipFont = nullptr);
+IMGUI_API void HelpMarker(std::function<const char* ()> getText, float width = 450.0f, ImFont* tooltipFont = nullptr);
```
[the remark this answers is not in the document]
SirCabby: `IMGUI_API void HelpMarker(const std::function<const std::string ()>* getText, float width = 450.0f, ImFont* tooltipFont = nullptr);` ?

src/plugins/map/MQ2Map.h
SirCabby: I did review all usage locations when making this change and I think it handles them appropriately.  I figured this was a QOL update to allow for managing them in the future a tad easier.  I'm happy to revert it if you're unsure about the safety of the change though.

src/imgui/ImGuiUtils.h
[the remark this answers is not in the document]
brainiac: not, const _reference_

const foo&

src/plugins/map/MQ2Map.h
brainiac: The implementation of it is supposed to be a static array of options, with a sentinel that terminates. could use std::array instead, but in either case changing it means the terminator is unnecessary, and so that should be changed as well.

src/plugins/map/MQ2Map.cpp
```diff
@@ -42,6 +42,8 @@ bool HighlightPulseIncreasing = true;
 int HighlightPulseIndex = 0;
 int HighlightPulseDiff = HighlightSIDELEN / 10;
 extern MapObject* gpActiveMapObjects;
+std::vector<MapFilterOption*> mapfilterObjectOptions;
+std::vector<MapFilterOption*> mapFilterOptions;
```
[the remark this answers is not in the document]
dannuic: since they aren't extern, I might suggest prefixing with `s_` to indicate that these are static maps. Also kind of weird that one variable is mapfilter and the other is mapFilter

src/plugins/map/MQ2Map.h
SirCabby: There's one spot where I look at the terminator still, whereas I could likely swap it out for just looking at the new vector size.  The other use case would be casting.  We have lots of static_cast, and in theory some of those casts could result in the Invalid option, which is used in places to signify that there is not a parent requirement etc.  Would you like me to remove it and exercise this plugin to ensure things still work, or would you like the option of casting to invalid to be left in place?

src/plugins/map/MQ2MapCommands.cpp
SirCabby: this was before making the MapFilterOptions a vector I believe.

src/plugins/map/MQ2Map.h
brainiac: I don't see how casting is relevant to this particular point. Can you show me an example?

Vector shouldn't be used because this is a static data set. it shouldn't be mutated and the number of elements is known at compile time. that makes it a prime candidate for std::array, but i would _not_ make it a vector. array has size(). But knowing the size means we don't need the terminator. The purpose of the terminator was specifically because we didn't know the size.

src/plugins/map/MQ2Map.cpp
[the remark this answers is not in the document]
brainiac: true, should be s_ and static if thats the case.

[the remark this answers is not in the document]
dannuic: I am actually wrong, I just found where it's extern'ed later and used in other implementation files.

src/plugins/map/MQ2Map.h
SirCabby: I wanted the vector because it makes it simpler to add new filter options in the future without mucking with a size on init.  But it's a good point that it's constant.  So you're suggesting to move it back to how it was without any modifications?  Or are you asking me to leverage the size and remove the terminator as well?

src/plugins/map/MQ2Map.cpp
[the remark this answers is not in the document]
SirCabby: good point dan, and the latter part was a typo that I didn't notice.

src/plugins/map/MQ2Map.h
brainiac: it would be less work to just change it back. I can convert it to an array later.

src/plugins/map/MQ2Map.cpp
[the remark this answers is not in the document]
SirCabby: oh, these are extern, does that change your naming suggestion?

src/plugins/map/MQ2Map.h
SirCabby: for the example, we use this method:
inline MapFilterOption& GetMapFilterOption(MapFilter Option)
{
    if (Option < MapFilter::All || Option >= MapFilter::Last)
        return MapFilterInvalidOption;

    return MapFilterOptions[static_cast<size_t>(Option)];
}

brainiac: Yeah you wouldn't change that.

brainiac: I'm not saying that you'd remove MapFilterInvalidOption, i'm saying you'd remove the instance of it at the end of the array.

src/plugins/map/MQ2MapCommands.cpp
SirCabby: my workarounds for getting some for of easily concat-able string into a non-const char* has been met with turmoil haha.  Is there a clean way to achieve this before I do a ton of conversions to make it work without _strdup?

brainiac: no, you need to change MapSetLocationCmd to take const char*

SirCabby: This is a little weird because MapSetLocationCmd is a registered plugin command.
`void AddCommand(const char* Command, fEQCommand Function`
an fEQCommand requires non-const. 
I could maybe workaround this by creating a passthrough method just to register as an fEQCommand and then call the current method with the const signature?

SirCabby: `using fEQCommand             = void   (*)(SPAWNINFO*, char* Buffer);`

brainiac: hmm i thought we had an overload for const char* too..

brainiac: you'll have to do this jank to get a char* from a string then: `&string[0]`

src/plugins/map/MQ2Map.h
SirCabby: I think this is the last unaddressed comment in the PR.  Were you wanting to change it to an array later as mentioned or leave as is?  I can revert a portion of this commit if you prefer it unchanged: https://github.com/SirCabby/macroquest/commit/9668f0e9dd144e119fc692c116ecdada641d0b73

src/plugins/map/MQ2Map.cpp
```diff
@@ -42,8 +42,8 @@ bool HighlightPulseIncreasing = true;
 int HighlightPulseIndex = 0;
 int HighlightPulseDiff = HighlightSIDELEN / 10;
 extern MapObject* gpActiveMapObjects;
-std::vector<MapFilterOption*> mapfilterObjectOptions;
-std::vector<MapFilterOption*> mapFilterOptions;
+std::vector<MapFilterOption*> s_mapFilterObjectOptions;
+std::vector<MapFilterOption*> s_mapFilterOptions;
```
brainiac, line 46: if these are prefixed with `s_` then they should also be static.

brainiac: the s_ means static or "file local"

src/plugins/map/MQ2Map.h
```diff
@@ -186,28 +186,28 @@ void RemoveGroundItem(GROUNDITEM* pGroundItem);
...
-	if (Option == MapFilter::Invalid)
-		return true;
+    if (Option == MapFilter::Invalid)
+        return true;
 
-	MapFilterOption& option = GetMapFilterOption(Option);
-	return IsOptionEnabled(option.RequiresOption);
+    MapFilterOption& option = GetMapFilterOption(Option);
+    return IsOptionEnabled(option.RequiresOption);
 }
```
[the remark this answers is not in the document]
SirCabby: I removed all the tabs in the file.  Would you prefer I only fix the tab I added?

src/plugins/map/MQ2Map.h
brainiac: just leave it, i'll address it.

src/plugins/map/MQ2Map.h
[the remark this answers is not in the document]
brainiac: read the original comment again, the issue was specifically with alignment within the line, not the indentation of the line.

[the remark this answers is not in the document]
brainiac: the soruce code is indented using tabs, yuo should not change that.

src/plugins/map/MQ2Map.h
```diff
@@ -136,8 +136,8 @@ extern char MapLeftClickString[MAX_CLICK_STRINGS][MAX_STRING];
 extern bool repeatMapshow;
 extern bool repeatMaphide;
 
-extern std::vector<MapFilterOption*> s_mapFilterObjectOptions;
-extern std::vector<MapFilterOption*> s_mapFilterOptions;
+static std::vector<MapFilterOption*> s_mapFilterObjectOptions;
+static std::vector<MapFilterOption*> s_mapFilterOptions;
```
brainiac, line 140: ok, so doing this is oging to put a copy of this variable in every source file. i'm pretty sure thats not what you want to do.

src/plugins/map/MapObject.cpp
SirCabby: finding the template from the maploc can throw (not really).  If I try / catch that though, it means this might not get reset.

SirCabby:
```
MapObjectMapLoc::~MapObjectMapLoc()
{
	try
	{
		auto thisTemplate = GetMapLocByTag(m_tag);
		thisTemplate->ClearReferenceToMapLoc();
	}
	catch (...)
	{
	}
	RemoveMapLoc();
}
```

brainiac: why would it throw?

SirCabby: my mistake, was tired.  This can be nullptr, not throw.  I'll fix heh

src/plugins/map/MapObject.cpp
SirCabby: so I made these changes, but it now throws in the mapobjectmaploc dtor when trying to look up the template to deref which is mid deletion as well.  I can maybe work around this but there's more going on than just these tweaks I think

src/plugins/map/MQ2MapCommands.cpp
SirCabby: I think I stole this from the imgui example haha, but I'll fix it

src/plugins/map/MQ2MapCommands.cpp
SirCabby: this was our last solution to removing the memory leak.  Do I just need to assign it first or something?

brainiac (approved):

PR #662, opened 2022-11-16
Added mq.pickle(file, table) command to lua

dannuic:


src/plugins/lua/bindings/lua_MQBindings.cpp
```diff
@@ -574,6 +574,113 @@ static sol::table lua_getFilteredSpawns(sol::this_state L, std::optional<sol::fu
...
+
+	std::error_code ec;
+	std::filesystem::create_directories(path.parent_path(), ec);
+	if (ec)
+	{
+		LuaError("Failed to create directory for pickling %.*s with error: %s", file_path.size(), file_path.data(), ec.message());
+		return;
+	}
+
+	std::ofstream ofs(path, std::ios_base::out | std::ios_base::trunc);
```
brainiac, line 676: try/catch, ofstream can throw

PR #671, opened 2022-12-01
Added lua actor implementation

dannuic:
Added lua actor implementation for communication between scripts (and lua plugins).


src/plugins/lua/LuaThread.cpp
```diff
@@ -215,6 +216,16 @@ int LuaThread::PackageLoader(const std::string& pkg, lua_State* L)
...
 	}
 
+	if (pkg == "actors")
+	{
+		LuaActors::RegisterLua(m_actorTable, sv);
+		m_globalState.set("_mq_internal_actors", *m_actorTable);
+
+		std::string_view script("return _mq_internal_actors");
+		luaL_loadbuffer(sv, script.data(), script.size(), pkg.c_str());
+		return 1;
```
brainiac, line 226: this can be simplified to something like the following. Your function just needs to return a table.

```
	if (pkg == "mq")
	{
		sol::stack::push(L, [this](sol::this_state L) { return RegisterMQNamespace(L); });
		return 1;
	}
```

brainiac: I revised the example because push doesn't properly identify the type of a lambda

dannuic: done

brainiac (approved):

PR #674, opened 2022-12-12
Setting up actor system using pipes

dannuic:
Adds protobuf for message protocols
Routes messages in the pipe server to processes
Routes messages in the pipe client to mailboxes

setup autologin to use the new system to demonstrate viability


include/mq/proto/ProtoPipes.h
```diff
@@ -0,0 +1,119 @@
...
+	T& Parse(T& obj)
+	{
+		return Parse(shared_from_this(), obj);
+	}
+
+	template <typename T>
+	static T Parse(const PipeMessagePtr& message)
+	{
+		T obj;
+		return Parse(message, obj);
```
brainiac, line 47: this too returns by copy.

brainiac: if you do
```
T obj;
Parse(message, obj);
return obj;
```
then it would rvo

include/mq/proto/ProtoPipes.h
```diff
@@ -0,0 +1,119 @@
...
+	}
+
+	template <typename ID, typename T>
+	static void SendProtoReply(const PipeMessagePtr& message, ID messageId, const T& obj, uint8_t status = 0)
+	{
+		std::string data = obj.SerializeAsString();
+		message->SendReply(static_cast<MQMessageId>(messageId), &data[0], data.size(), status);
+	}
+
+	const std::optional<proto::Address>& GetReturn() { return m_returnAddress; }
```
brainiac, line 70: GetSender? GetReturnAddress?

`Return` doesn't feel like the proper term to describe what this is.

src/main/MQ2PostOffice.h
```diff
@@ -23,19 +24,30 @@ void ShutdownPipeClient();
...
+}
 
+MQLIB_OBJECT void RouteMessage(MQMessageId messageId, const std::string& data);
+template <typename ID>
+void RouteMessage(ID messageId, const std::string& data)
+{
+	RouteMessage(static_cast<MQMessageId>(messageId), data);
+}
+
+MQLIB_OBJECT void RouteMessage(const std::string& data); // envelopes come as strings, and will always want "ROUTE" message IDs
```
brainiac, line 49: if this is the public api, it should be somewhere in include/mq/ and have comments describing each

src/autologin/AutoLogin.cpp
```diff
@@ -14,7 +14,9 @@
 
 // AutoLogin code shared between MacroQuest.exe and MQ2AutoLogin.dll
 
-#include "AutoLoginShared.h"
+#include "AutoLogin.h"
```
brainiac, line 17: src/autologin is easy to confuse with src/plugins/autologin

please give this folder a name identifying that it is the shared component.

brainiac (commented):
left some comments on stuff and things

include/mq/proto/Mailbox.h
```diff
@@ -0,0 +1,218 @@
...
+				unwrapped->GetHeader()->messageId = static_cast<MQMessageId>(envelope.message_id());
+
+			if (envelope.has_return_address())
+				unwrapped->SetReturn(envelope.return_address());
+
+			return unwrapped;
+		}
+
+		template <typename T>
+		std::string Stuff(const proto::Address& address, MQMessageId messageId, const T& obj)
```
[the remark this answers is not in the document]
dannuic: It's stuffing an envelope. That's what it's called when you put a letter into it.

include/mq/proto/ProtoPipes.h
```diff
@@ -0,0 +1,119 @@
...
+class ProtoMessage : public PipeMessage
+{
+public:
+	using PipeMessage::PipeMessage;
+	
+	template <typename T>
+	T Parse()
+	{
+		T obj;
+		return Parse(shared_from_this(), obj);
```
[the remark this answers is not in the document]
dannuic: This shouldn't be doing that anymore

include/mq/proto/ProtoPipes.h
dannuic: This shouldn't be doing that anymore

src/autologin/AutoLogin.cpp
dannuic: Changed the name to login

include/mq/proto/Mailbox.h
```diff
@@ -0,0 +1,219 @@
...
+				PostReply(message, *returnAddress, messageId, obj, status);
+			}
+			else
+			{
+				message->SendProtoReply(messageId, obj, status);
+			}
+		}
+
+		template <typename ID, typename T>
+		void PostReply(const PipeMessagePtr& message, const proto::Address& returnAddress, ID messageId, const T& obj, uint8_t status = 0)
```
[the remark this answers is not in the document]
dannuic: 2 things that templating here gets us:
first, it allows us to set an ID type (that can be cast as uint32_t) that the individual post offices understand to ensure we aren't sending messages with invalid ID's, and to prevent us from needing to constantly cast to uint32_t at the call site.
second, it allows us to have actual serializable objects without needing to .SerializeAsString() at the call site.

This is a tiny amount of overhead in terms of multiple functions in the compiled code when you compare it against the convenience of the interface provided to the developer of new post offices, IMO.

include/mq/proto/Mailbox.h
```diff
@@ -0,0 +1,219 @@
...
+				unwrapped->GetHeader()->messageId = static_cast<MQMessageId>(envelope.message_id());
+
+			if (envelope.has_return_address())
+				unwrapped->SetReturn(envelope.return_address());
+
+			return unwrapped;
+		}
+
+		template <typename ID, typename T>
+		std::string Stuff(const proto::Address& address, ID messageId, const T& obj)
```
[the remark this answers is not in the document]
dannuic: same as above

src/main/MQ2PostOffice.h
dannuic: Hmm, this is actually a good point. I think I originally intended for this to be for managing mailboxes in the MQ Main post office, but I think the line is muddy now and the MQ main post office should be the only post office in MQ (the launcher would have the other post office currently, and other apps would have other post offices)

include/mq/proto/Mailbox.h
```diff
@@ -0,0 +1,219 @@
...
+#include <queue>
+#include <memory>
+
+namespace mq {
+namespace mailbox {
+
+// we should assume that everything lives inside an Envelope here. All mail must be
+// in an envelope, no postcards (yet), but we open the Envelope to create ProtoMessages
+// when we put them on the queue
+class PostOffice
```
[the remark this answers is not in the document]
dannuic: changed it

include/mq/proto/Mailbox.h
```diff
@@ -0,0 +1,219 @@
...
+
+	public:
+		Mailbox(
+			std::string_view localAddress,
+			ReceiveCallback receive,
+			PostCallback post
+		)
+			: m_localAddress(localAddress)
+			, m_receive(receive)
+			, m_post(post)
```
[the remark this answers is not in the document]
dannuic: changed to const ref

include/mq/proto/Mailbox.h
```diff
@@ -0,0 +1,218 @@
...
+	PostOffice(void(*postAction)(const std::string&))
+		: m_post(postAction)
+	{}
+
+	std::shared_ptr<Mailbox> CreateMailbox(const std::string& localAddress, ReceiveCallback receive)
+	{
+		return std::make_shared<Mailbox>(localAddress, receive, m_post);
+	}
+
+	bool AddMailbox(const std::string& localAddress, std::shared_ptr<Mailbox> mailbox)
```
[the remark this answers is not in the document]
dannuic: removed, was a relic from when I was moving the mailbox

include/mq/proto/ProtoPipes.h
dannuic: I like Get/SetSender, changed to that

src/main/MQ2PostOffice.h
dannuic: alright, I moved all this stuff to mq/proto/PostOffice.h -- there are still some leftover functions here (and in the launcher) that directly send messages, which is legacy support. We could move those to actual actors in the future.

brainiac (approved):

PR #676, opened 2023-01-05
MQ2Commands - fix /removeaug

### LINKED ISSUE #669: /removeaug does not seem to work correctly
Using:

```
/removeaug augid itemid
```
or
```
/removeaug augname itemid
```
or
```
/removeaug augid itemname
```
or
```
/removeaug augname itemname
```

I am wondering if it is looking for a perfect distiller? I realize that previously some variables/values were switched but those should be resolved?

I have both the max distiller (under perfect) and the level appropriate ones and still no dice.

Sicprofundus:
- resolves #669 
- replaced pLocalPC->GetItemByID with FindItemByID
- this fixes /removeaug "aug name" "item"

### DIFF
--- a/src/main/MQ2Commands.cpp
+++ b/src/main/MQ2Commands.cpp
@@ -5696,7 +5696,8 @@ static ItemPtr FindAugmentSolvent(const ItemPtr& pAugItem)
 	{
 		int realID = pDistillerInfo.GetIDFromRecordNum(i, false);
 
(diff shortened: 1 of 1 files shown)


brainiac (changes requested):
missed the other side of the else case, and you also left the previous code commented out.

don't worry about it though, i'm taking care of it.

brainiac:
Thanks for identifying the issue, btw.

Issue #36, opened 2019-12-29
ItemDB.txt - Update Method


MacroQuestGitHubMigrationBot:
In GitLab by @Knightly1 on Dec 29, 2019, 10:15

ItemDB.txt is missing quite a few items.  Either update the methodology for this or find a way to maintain it easier.

brainiac:
The refactor of mq2linkdb makes this irrelevant

PR #678, opened 2023-01-09
Implements a Set / Remove Loot Filter Command

Ortster:
- Implements /setlootfilter, used as /setlootfilter (option) \"itemID^itemIconID^itemName\"  
Option being AN|AG|NV|AR|ANAR|AGAR|NVAR, and the item portion being the line needed in the advloot inis.  
For example: /setlootfilter ANAR "13072^793^Rat Ears", setting Rat Ears to Always Need and Always Roll  
On use will add/set the advloot filter for the item given, with the option given.  Sets options to what was set, removes any previous options.  For example using AN will only set Always Need, Always Greed/Never/Always Roll will get removed.

- Implements /removelootfilter, used as /removelootfilter itemID  
For example: /removelootfilter 13072, which removes the advloot filter for Rat Ears
On use will remove the advloot filter for the item with the given item ID.  

Main reasoning for the commands is to simplify the process of setting advloot filters.  Avoids having to manually edit ini files, or situations in game where you may not have the item or its item link available to set the filters through the normal method. Also allows the user to broadcast one of these commands to modify advloot filters for multiple characters at the same time.

### DIFF
--- a/src/main/MQ2CommandAPI.cpp
+++ b/src/main/MQ2CommandAPI.cpp
@@ -962,6 +962,8 @@ void InitializeMQ2Commands()
 		{ "/removeaura",        RemoveAura,                 false, true  },
 #if HAS_ADVANCED_LOOT
 		{ "/advloot",           AdvLootCmd,                 true,  true  },
+		{ "/setlootfilter",     SetLootFilterCmd,           true,  true  },
+		{ "/removelootfilter",  RemoveLootFilterCmd,        true,  true  },
 #endif
 		{ "/pickzone",          PickZoneCmd,                true,  true  },
 		{ "/assist",            AssistCmd,                  true,  true  },
--- a/src/main/MQ2Commands.cpp
+++ b/src/main/MQ2Commands.cpp
@@ -4944,6 +4944,167 @@ void AdvLootCmd(SPAWNINFO* pChar, char* szLine)
 		cmdAdvLoot(pChar, szLine);
 	}
 }
+
+// ***************************************************************************
+// Function:    SetLootFilterCmd
+// Description: '/setlootfilter' command
+//              Adds the ability to set advanced loot filters via command line
+// Usage: /setlootfilter AN|AG|NV|AR|ANAR|AGAR|NVAR "itemID^itemIconID^itemName"
+// ***************************************************************************
+void SetLootFilterCmd(SPAWNINFO* pChar, char* szLine)
+{
+	if (pLootFiltersManager)
+	{
+		// No arguments, display usage
+		if (!szLine[0])
+		{
+			WriteChatf("Usage: /setlootfilter (option) \"itemID^itemIconID^itemName\"");
+			WriteChatf("Options:");
+			WriteChatf("       AN - Always Need");
+			WriteChatf("       AG - Always Greed");
+			WriteChatf("       NV - Never");
+			WriteChatf("       AR - Always Roll");
+			WriteChatf("       ANAR - Always Need and Always Roll");
+			WriteChatf("       AGAR - Always Need and Always Greed");
+			WriteChatf("       NVAR - Never and Always Roll");
+			WriteChatf("Example: /setlootfilter ANAR \"13072^793^Rat Ears\"");
+			WriteChatf("       Sets AdvLoot filter for \"Rat Ears\" to Always Need and Always Roll");
+			return;
+		}
+
+		// Grab user's selected filter type
+		char filterTypeArg[MAX_STRING] = { 0 };
+		GetArg(filterTypeArg, szLine, 1);
+
+		// Blank filter types, set by user below
+		int filterTypes = 0;
+
+		// Always Need
+		if (_stricmp(filterTypeArg, "AN") == 0)
+		{
+			filterTypes |= LootFilterBit(LootFilterType_AlwaysNeed);
+		}
+		// Always Greed
+		else if (_stricmp(filterTypeArg, "AG") == 0)
+		{
+			filterTypes |= LootFilterBit(LootFilterType_AlwaysGreed);
+		}
+		// Never
+		else if (_stricmp(filterTypeArg, "NV") == 0)
+		{
+			filterTypes |= LootFilterBit(LootFilterType_NeverLoot);
+		}
+		// Auto Roll
+		else if (_stricmp(filterTypeArg, "AR") == 0)
+		{
+			filterTypes |= LootFilterBit(LootFilterType_AutoRoll);
+		}
+		// Always Need / Auto Roll
+		else if (_stricmp(filterTypeArg, "ANAR") == 0)
+		{
+			filterTypes |= LootFilterBit(LootFilterType_AlwaysNeed);
+			filterTypes |= LootFilterBit(LootFilterType_AutoRoll);
+		}
+		// Always Greed / Auto Roll
+		else if (_stricmp(filterTypeArg, "AGAR") == 0)
+		{
+			filterTypes |= LootFilterBit(LootFilterType_AlwaysGreed);
+			filterTypes |= LootFilterBit(LootFilterType_AutoRoll);
+		}
+		// Never / Auto Roll
+		else if (_stricmp(filterTypeArg, "NVAR") == 0)
+		{
+			filterTypes |= LootFilterBit(LootFilterType_NeverLoot);
+			filterTypes |= LootFilterBit(LootFilterType_AutoRoll);
+		}
+		else
+		{
+			WriteChatf("Invalid Filter Type Entered, AN / AG / NV / AR / ANAR / AGAR / NVAR Required.");
+			return;
+		}
+
+		// We properly set a filter type from argument, grab item filter string from rest
+		if (filterTypes != 0)
+		{
+			// Grab item filter argument from user
+			char itemFilterArg[MAX_STRING] = { 0 };
+			GetMaybeQuotedArg(itemFilterArg, MAX_STRING, szLine, 2);
+
+			if (itemFilterArg == nullptr)
+			{
+				WriteChatf("Invalid Item Filter Argument.");
+			}
+
+			// Split out each item in the item filter argument
+			std::vector<std::string> itemFilterStrSplit;
+			std::stringstream ss(itemFilterArg);
+			std::string item;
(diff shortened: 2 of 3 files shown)


Knightly1:
This idea is pretty good, but I'm a little concerned about the implementation specifics.  The namespace first, using two commands where it might be better to use one...but also I'm not a fan of having to enter in all of the information for the item.  And even if you are going to enter all the information for the item, I'm not a fan of entering it delimited by carats.

I'll have a look at the code, but those are my initial thoughts from the summary.

Ortster:
Ideally you'd just pass in the item ID, was not sure if there is a way to get the other two pieces of information from the EQ client from just having the itemID.  ItemDisplay does the same thing this is doing but has access to those three pieces of information easily.

The idea with going for the delimited by carats is just that it matches how the item gets added into the ini(s). I figured most use cases the user is on a site (e.g. https://www.redguides.com/community/advloot-gen) to get the line for the item and just pastes that right into the command. There may be some use cases where the user is grabbing that info off ItemDisplay, which I agree the carats are a bit clunky at that point.

Merging the two commands might be a good idea, even better if there is a way to cut it down to just itemID on both.  I separated them out just for an easier/quicker/cleaner implementation. 


This is somewhat of a re-implementation of a command that looks like it used to exist? Or at least there is something similar in the changelog. I didn't dig too much into ItemDisplay to see if it was in there but renamed. Had the same limitation with requiring all three pieces of information, and for some reason hard set to Always Greed / Always Roll.

> Jan 15 2018 by eqmule
> - Added `/AddLootFilter syntax:  /AddLootFilter <itemID> <ItemIconID> \"Item Name\"`
>   Example: `/AddLootFilter ${Cursor.ID} ${Cursor.IconID} "${Cursor.Name}"`
>   will add whateveris on your cursor to the advloot filters as always greed and roll.
> - Added a AddLootFilter method to the DisplayItem TLO.
> - Usage: `/invoke ${DisplayItem.AddLootFilter} or ${DisplayItem[x].AddLootFilter}` where x is 1 to 6
> - this will add whatever item is in the DisplayItem to advloot filters as AG and Roll


Take a look at it and let me know if the general concept is sound. Ill take a look shortly at merging the two commands into one and changing the arguments around to be split out. If you know any magic to pull IconID and Name from the EQ Client when you just have the ItemID I am all ears, as that cleans this up quite a bit!

src/main/MQ2Commands.cpp
```diff
@@ -4944,6 +4944,167 @@ void AdvLootCmd(SPAWNINFO* pChar, char* szLine)
...
+
+		// Grab user's selected filter type
+		char filterTypeArg[MAX_STRING] = { 0 };
+		GetArg(filterTypeArg, szLine, 1);
+
+		// Blank filter types, set by user below
+		int filterTypes = 0;
+
+		// Always Need
+		if (_stricmp(filterTypeArg, "AN") == 0)
```
dannuic, line 4983: don't use stricmp, use the string utils (`ci_equals`)

src/main/MQ2Commands.cpp
```diff
@@ -4944,6 +4944,167 @@ void AdvLootCmd(SPAWNINFO* pChar, char* szLine)
...
+			if (itemFilterStrSplit.size() != 3)
+			{
+				WriteChatf("Invalid Item Filter Argument. Check Formatting.");
+				return;
+			}
+
+			// Grab out ItemID, Item IconID, and ItemName
+			int itemID = GetIntFromString(itemFilterStrSplit.at(0), -1);
+			int itemIconID = GetIntFromString(itemFilterStrSplit.at(1), -1);
+			std::string itemName = itemFilterStrSplit.at(2);
```
dannuic, line 5056: Two things here:
- first, don't use carats. If someone is looking up the direct file modification string, they are going to directly modify the file. spaces are better.
- second, don't use stringstream. Use the string utils provided to you in mq (there are tokenization functions available).

As a bit of an aside. I'm pretty sure that item name here is case sensitive, so that's going to trip a lot of users up. I have a more general comment about that later.

src/main/MQ2Commands.cpp
```diff
@@ -4944,6 +4944,167 @@ void AdvLootCmd(SPAWNINFO* pChar, char* szLine)
...
+		// We properly set a filter type from argument, grab item filter string from rest
+		if (filterTypes != 0)
+		{
+			// Grab item filter argument from user
+			char itemFilterArg[MAX_STRING] = { 0 };
+			GetMaybeQuotedArg(itemFilterArg, MAX_STRING, szLine, 2);
+
+			if (itemFilterArg == nullptr)
+			{
+				WriteChatf("Invalid Item Filter Argument.");
```
dannuic, line 5035: you don't early out here, so you go on to try to parse a nullptr

src/main/MQ2CommandAPI.cpp
```diff
@@ -962,6 +962,8 @@ void InitializeMQ2Commands()
 		{ "/removeaura",        RemoveAura,                 false, true  },
 #if HAS_ADVANCED_LOOT
 		{ "/advloot",           AdvLootCmd,                 true,  true  },
+		{ "/setlootfilter",     SetLootFilterCmd,           true,  true  },
+		{ "/removelootfilter",  RemoveLootFilterCmd,        true,  true  },
```
dannuic, line 966: I suggest that you add commands ('addfilter' and 'removefilter' maybe?) to the `/advloot` command instead of creating two new commands.

dannuic (changes requested):
Some general notes here: The carats have been mentioned, but even worse is that people need so much information for this to work (I know you said that already). I don't really see this getting used much. What really needs to happen for this is that you would maintain a database of item IDs or figure out an API you can call with all the needed information so that a user could simply use the item ID on the command line. Or better yet, use item names with case insensitive searches that print out item IDs you can choose from in the case that the user needs to disambiguate.

This isn't a new need for mq, there are other things that would really benefit from being able to lookup item information from an external source, but barring that, requiring everything on the command line is fine for now.

TL;DR: requiring everything on the command line is fine until someone implements an item database tool.

Ortster:
Pushed some changes that cleans it up into one function that does both add/remove and a handful of the other suggestions.

Like I mentioned on the use case, main use is for multi boxers who want to set loot filters all at once across boxes without fiddling with item links or the inis. Kind of brought up by item links from the SoF expansion being a bit weird when obtained from MQ2LinkDB.

The database of IDs / API for grabbing item info would be great. It kind of dawned on me earlier that MQ2LinkDB actually sorta kinda has that. Albeit from item links from a text file, and whatever information they have inside.

src/main/MQ2Commands.cpp
```diff
@@ -4958,151 +4959,126 @@ void SetLootFilterCmd(SPAWNINFO* pChar, char* szLine)
...
+		{
+			// Grab Item ID
+			GetArg(szArg, szRest, 1);
+			szRest = GetNextArg(szRest, 1);
+
+			// Grab/Check Int of Item ID
+			int itemID = GetIntFromString(szArg, -1);
+			if (itemID == -1)
+			{
+				WriteChatf("Unable to remove AdvLoot Filter, Item ID Invalid.");
```
dannuic, line 4994: minor nitpick here, but it's usually better to include the text that it tried to parse in the error message

src/main/MQ2Commands.cpp
```diff
@@ -4958,151 +4959,126 @@ void SetLootFilterCmd(SPAWNINFO* pChar, char* szLine)
...
-		}
-		else
-		{
-			WriteChatf("Invalid ItemID Entered.");
+			WriteChatf("Unable to set AdvLoot Filter, Item Name Invalid.");
 			return;
 		}
+
+		// Call client function to set advloot filter
+		pLootFiltersManager->SetItemLootFilter(itemID, itemIconID, itemName.data(), filterTypes);
```
dannuic, line 5080: don't use .data() here, there's no guarantee that it's null terminated, use .c_str() -- but why are you even using a string here because it's directly just `szArg`? Simply use `szArg` and make sure it's not empty (`szArg[0] == 0` should suffice since that's the only validation you are doing anyway)

dannuic (changes requested):

src/main/MQ2Commands.cpp
Ortster: I am used to using QString out of Qt for string manipulation and such, so std::string is close. Mostly just a holdover from when I was initially implementing this.  Should be swapped over to using szArg with that empty check, and the description string now using .c_str().

src/main/MQ2Commands.cpp
Ortster: All the error messages should now print out what the argument was its complaining about.  Also missed the filter type argument having a check for it missing, which now has a message as well.

src/main/MQ2Commands.cpp
dannuic: The issue here is the you're doing an extra allocation and copy when you create the string, only to just convert it back into a c-string

Ortster: Yeah I get that, like I said just a holdover from when I was messing around with initial implementation.

Latest push should fix everything thats been mentioned.

Knightly1:
dannuic's TLDR sums up why I haven't merged this yet:

> TL;DR: requiring everything on the command line is fine until someone implements an item database tool.

This is true and no one really has an issue with this code contribution, I've just been hesitant to merge this since it is essentially just a shortcut to /ini at that point.

What I'm going to do is move the code into MQ2LinkDB under the LinkDB command for now.  That way you can do this by ItemID or Name without having to know the Icon ID and hopefully that will make things a bit friendlier.  We can move it into core under an appropriate command once we have a proper core item database which is something we'd like to do anyway (item #36)

In the meantime I'm going to leave this issue open on the core repo until I can link to the LinkDB work (normally I would close the issue and link to another issue in the other repo but it was good work by Ortster and I don't want it to get lost or deprioritized).

Ortster:
Didn't really think of putting it into MQ2LinkDB, would help quite a bit with the clunkiness of using the command.

Appreciate the effort on still implementing the concept. Niche use but does help multiboxers quite a bit when dealing with loot filters. No messing with inis or linking an item around to each character. Broadcast the filter command and you are good to go.

Feel free to close out whenever!

PR #686, opened 2023-01-17
Put lib/exp files in MQ2LibDir on release build.

mkolassa:
I find myself compiling this not only for myself but also for friends that play, and it's far more convenient to have these intermediate files placed elsewhere than it is to clean the directory after a build.  As an added benefit, with them going to the ./lib/ directory instead of the trash can, incremental builds are still available for faster compile times.

### DIFF
--- a/src/Common.props
+++ b/src/Common.props
@@ -72,6 +72,10 @@
       <OptimizeReferences>false</OptimizeReferences>
       <EnableCOMDATFolding>false</EnableCOMDATFolding>
     </Link>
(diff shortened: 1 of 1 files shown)


mkolassa:
I have removed the code that would have disabled the pdb file, so now this only moves exp/lib files into the ./build/lib/ directory.

mkolassa:
It appears that moving these files is not popular with the folks on Discord, so I guess this PR serves no purpose.

PR #695, opened 2023-02-17
Fix PackFind to return correctly empty pack slot or inventory slot

projecteon:
Fix PackFind to return correctly empty pack slot or inventory slot based on room for given item

### DIFF
--- a/include/moveitem.h
+++ b/include/moveitem.h
@@ -297,3 +297,44 @@ int FreeSlotForItem(ItemClient* pItem)
 	}
 	return 0;
 }
+
+CItemLocation* FreeItemLocationForItem(CItemLocation* pFreeSlot, ItemClient* pItemClient)
+{
+	unsigned char  ucCurSize = 10;
+	ItemDefinition* pUnequipInfo = pItemClient->GetItemDefinition();
+	if (!pUnequipInfo) {
+		return nullptr;
+	}
+
+	for (unsigned short usSlot = InvSlot_FirstBagSlot; usSlot <= GetHighestAvailableBagSlot(); usSlot++)
+	{
+		if (ItemClient* pInvSlot = GetPcProfile()->GetInventorySlot(usSlot))
+		{
+			ItemDefinition* pItemInfo = pInvSlot->GetItemDefinition();
+
+			if (pItemInfo
+				&& pItemInfo->Combine != 2
+				&& pInvSlot->IsContainer()
+				&& pUnequipInfo->Size <= pItemInfo->SizeCapacity)
+			{
+				for (unsigned short usPack = 0; usPack < pItemInfo->Slots; usPack++) {
+					if (!pInvSlot->GetHeldItem(usPack)) {
+						pFreeSlot->InvSlot = usSlot;
+						pFreeSlot->BagSlot = usPack;
+						pFreeSlot->pBagSlot = pInvSlot->GetHeldItem(usPack);
+						pFreeSlot->pInvSlot = pInvSlot;
+						return pFreeSlot;
+					}
+				}
+			}
+		}
+		else {
+			pFreeSlot->InvSlot = usSlot;
+			pFreeSlot->BagSlot = INVALID_PACK;
+			pFreeSlot->pBagSlot = nullptr;
+			pFreeSlot->pInvSlot = nullptr;
+			return pFreeSlot;
+		}
+	}
+	return nullptr;
+}
--- a/include/moveitem.h
+++ b/include/moveitem.h
@@ -302,26 +302,22 @@ CItemLocation* FreeItemLocationForItem(CItemLocation* pFreeSlot, ItemClient* pIt
 {
 	unsigned char  ucCurSize = 10;
 	ItemDefinition* pUnequipInfo = pItemClient->GetItemDefinition();
-	if (!pUnequipInfo) {
-		return nullptr;
-	}
 
 	for (unsigned short usSlot = InvSlot_FirstBagSlot; usSlot <= GetHighestAvailableBagSlot(); usSlot++)
 	{
 		if (ItemClient* pInvSlot = GetPcProfile()->GetInventorySlot(usSlot))
 		{
 			ItemDefinition* pItemInfo = pInvSlot->GetItemDefinition();
-
-			if (pItemInfo
-				&& pItemInfo->Combine != 2
+			if (pItemInfo->ContainerType != ContainerType_Quiver
 				&& pInvSlot->IsContainer()
 				&& pUnequipInfo->Size <= pItemInfo->SizeCapacity)
 			{
 				for (unsigned short usPack = 0; usPack < pItemInfo->Slots; usPack++) {
-					if (!pInvSlot->GetHeldItem(usPack)) {
+					auto pBagSlot = pInvSlot->GetHeldItem(usPack);
+					if (!pBagSlot) {
 						pFreeSlot->InvSlot = usSlot;
 						pFreeSlot->BagSlot = usPack;
-						pFreeSlot->pBagSlot = pInvSlot->GetHeldItem(usPack);
+						pFreeSlot->pBagSlot = pBagSlot;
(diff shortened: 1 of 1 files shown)


include/moveitem.h
```diff
@@ -189,57 +189,40 @@ inline int CountItem(const char* pcItemName, int usMin, int usInvSlots)
...
-
-// finds an open slot in a bag that has items (not empty bags)
+// finds an open slot in a bag or inventory
 CItemLocation* PackFind(CItemLocation* pFreeSlot, ItemClient* pUnequipSlot)
 {
 	unsigned char  ucCurSize = 10;
-	unsigned short usSlot    = 0;
-
-	unsigned short usDecrement = 0;
+	ItemDefinition* pUnequipInfo = pUnequipSlot->GetItemDefinition();
```
dannuic, line 196: If this is null, you can avoid the loop instead of checking if it's null in every iteration

include/moveitem.h
```diff
@@ -189,57 +189,40 @@ inline int CountItem(const char* pcItemName, int usMin, int usInvSlots)
...
-				pLAST = nullptr;
 			}
 		}
+		else {
+			pFreeSlot->InvSlot = usSlot;
+			pFreeSlot->BagSlot = INVALID_PACK;
+			pFreeSlot->pBagSlot = nullptr;
+			pFreeSlot->pInvSlot = nullptr;
+			return pFreeSlot;
+		}
```
dannuic, line 225: I don't think you should return an invalid slot on _any_ failure. You could still have valid options later in the loop, no?

include/moveitem.h
```diff
@@ -189,57 +189,40 @@ inline int CountItem(const char* pcItemName, int usMin, int usInvSlots)
...
 
-	// no user-specific slot given, find the first free bag slot
-	for (usSlot = InvSlot_FirstBagSlot; usSlot <= GetHighestAvailableBagSlot(); usSlot++)
+	for (unsigned short usSlot = InvSlot_FirstBagSlot; usSlot <= GetHighestAvailableBagSlot(); usSlot++)
 	{
 		if (ItemClient* pInvSlot = GetPcProfile()->GetInventorySlot(usSlot))
 		{
-			ItemDefinition* pItemInfo    = GetItemFromContents(pInvSlot);
-			ItemDefinition* pUnequipInfo = GetItemFromContents(pUnequipSlot);
+			ItemDefinition* pItemInfo = pInvSlot->GetItemDefinition();
```
dannuic, line 202: looking at this logic here: we would fail to return open base inventory slots, no? I know that's how the old algorithm does it as well, but that doesn't seem correct

include/moveitem.h
```diff
@@ -189,57 +189,40 @@ inline int CountItem(const char* pcItemName, int usMin, int usInvSlots)
...
-						}
-					} else if (usDecrement) {
-						//DebugSpewAlways("pack %hu minus decrement %hu equals %hu", usPack, usDecrement, usPack-usDecrement);
-						pFreeSlot->InvSlot    = usSlot;
-						pFreeSlot->BagSlot    = usPack;
-						pFreeSlot->pBagSlot   = pInvSlot->GetHeldItem(usPack - usDecrement); // will be NULL
-						pFreeSlot->pInvSlot   = pInvSlot;
+						pFreeSlot->InvSlot = usSlot;
+						pFreeSlot->BagSlot = usPack;
+						pFreeSlot->pBagSlot = pInvSlot->GetHeldItem(usPack);
```
dannuic, line 212: why not just explicitly set this as nullptr? You know that's what it is, because that was the check

include/moveitem.h
projecteon: What do you mean by this?
This is the case where the slot is not a container and is empty, which means it can fit any item including the incoming one so its a valid slot to put the item.

include/moveitem.h
projecteon: No this should be caught by the code in the previous comment.

include/moveitem.h
projecteon: I never debugged the code and actually checked it was a nullptr, this was left from the old code and I didnt dare change it.

projecteon:
Issue/usecase @brainiac 

MQ2Exchange `/unequip slot` does not work unless you have empty slot in the first bag in your inventory. Expect it to un-equip as along as it finds a valid slot in your inventory (main slot or space in a big enough bag)

Debugging this into moveitem.h PackFind which seems to not search all inventory slots (fails if first inventory slot has a bag and its full)
Given the comment on this function I assume this refactor is wrong and the usage of it has been wrong. MQ2Melee also uses this function to handle `unequip` which probably doesnt work.

The following public plugins use this file:
MQ2Bandolier.cpp
MQ2Cursor.cpp
MQ2Exchange.cpp
MQ2Melee.cpp

MQ2Bandolier and MQ2Cursor does not use the `PackFind` function. MQ2Bandolier seems to make its own functions to handle exactly this functionality.

projecteon:
Decided to rework this change request.
Added a function that can be used in MoveUtils that actually finds a valid empty spot in either main inventory or inside a back. 
If this gets approved Ill make a PR to MoveUtils to use the new function.

projecteon:
Had another think around this.. 
Is it more preferable to pull this out of MQ core and do as MQ2Bandolier and MQ2Cursor and just implement this locally in affected plugins?

include/moveitem.h
```diff
@@ -297,3 +297,44 @@ int FreeSlotForItem(ItemClient* pItem)
...
 	return 0;
 }
+
+CItemLocation* FreeItemLocationForItem(CItemLocation* pFreeSlot, ItemClient* pItemClient)
+{
+	unsigned char  ucCurSize = 10;
+	ItemDefinition* pUnequipInfo = pItemClient->GetItemDefinition();
+	if (!pUnequipInfo) {
+		return nullptr;
+	}
```
brainiac, line 307: if pItemClient is non-null, then GetItemDefinition will always return non-null

Its not possible for an item to have no definition.

include/moveitem.h
```diff
@@ -297,3 +297,44 @@ int FreeSlotForItem(ItemClient* pItem)
...
+
+	for (unsigned short usSlot = InvSlot_FirstBagSlot; usSlot <= GetHighestAvailableBagSlot(); usSlot++)
+	{
+		if (ItemClient* pInvSlot = GetPcProfile()->GetInventorySlot(usSlot))
+		{
+			ItemDefinition* pItemInfo = pInvSlot->GetItemDefinition();
+
+			if (pItemInfo
+				&& pItemInfo->Combine != 2
```
brainiac, line 316: Use `ContainerType` instead of `Combine`.

2 maps to ContainerType_Quiver, so use that instead of 2.

include/moveitem.h
```diff
@@ -297,3 +297,44 @@ int FreeSlotForItem(ItemClient* pItem)
...
+				&& pInvSlot->IsContainer()
+				&& pUnequipInfo->Size <= pItemInfo->SizeCapacity)
+			{
+				for (unsigned short usPack = 0; usPack < pItemInfo->Slots; usPack++) {
+					if (!pInvSlot->GetHeldItem(usPack)) {
```
brainiac, line 321: This is doing a container look up, and it happens again right below, so save off the value so you don't need to look it up twice.

include/moveitem.h
```diff
@@ -297,3 +297,44 @@ int FreeSlotForItem(ItemClient* pItem)
...
+		return nullptr;
+	}
+
+	for (unsigned short usSlot = InvSlot_FirstBagSlot; usSlot <= GetHighestAvailableBagSlot(); usSlot++)
+	{
+		if (ItemClient* pInvSlot = GetPcProfile()->GetInventorySlot(usSlot))
+		{
+			ItemDefinition* pItemInfo = pInvSlot->GetItemDefinition();
+
+			if (pItemInfo
```
brainiac, line 315: same here with pItemInfo never being null

brainiac (changes requested):

brainiac (approved):

PR #696, opened 2023-02-17
Add GM check to return max available slots

projecteon:


src/main/MQ2Utilities.cpp
```diff
@@ -5154,6 +5154,9 @@ int GetHighestAvailableBagSlot()
 
 	int highestInvSlot = InvSlot_LastBonusBagSlot;
 
+	if (pLocalPC->bGM)
+		return highestInvSlot;
```
brainiac, line 5158: Can you add a comment that explains why we care about GM flag (its for emu?)

projecteon: Hope that was good enough

PR #700, opened 2023-02-24
Remove windows.h from mq/base/String.h

Knightly1:
- Replace references to codecvt
- Remove Common.h from String.h since it includes String.h since there are no dependencies
- Move utf8_to_wstring and wstring_to_utf8 into their own implementation file

### DIFF
--- a/include/mq/base/String.h
+++ b/include/mq/base/String.h
@@ -14,8 +14,6 @@
 
 #pragma once
 
-#include <mq/base/Common.h>
-
 #include <algorithm>
 #include <charconv>
 #include <string>
@@ -25,8 +23,6 @@
 #include <numeric>
 #include <unordered_map>
 #include <unordered_set>
-
-#include <Windows.h>
 
 namespace mq {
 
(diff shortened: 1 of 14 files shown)


include/mq/base/String.h
```diff
@@ -694,31 +690,7 @@ inline bool GetBoolFromString(const std::string_view svString, const bool defaul
...
-		return {};
-
-	int sizeNeeded = ::WideCharToMultiByte(CP_UTF8, 0, &s[0], (int)s.size(), nullptr, 0, nullptr, nullptr);
-	std::string r;
-	r.resize(sizeNeeded, 0);
-	::WideCharToMultiByte(CP_UTF8, 0, s.c_str(), (int)s.size(), &r[0], sizeNeeded, nullptr, nullptr);
-	return r;
-}
+std::wstring utf8_to_wstring(std::string_view s);
+std::string wstring_to_utf8(const std::wstring& s);
```
brainiac, line 694: should these be exported since they're implemented in mq2main?

Knightly1: I'll defer to your expertise here, but I think in this case no -- the intent here is to leave the implementation to whatever is bringing in this file (or to have no implementation for those at all if they're not needed).  In this case, MQ2Main is using WStringUtils.cpp to handle the implementation, and MacroQuest.exe does the same (by bringing in that file) but it's not intended to be used like an actual library (I didn't separate it for that reason) since the implementation will differ on non-windows platforms.

brainiac: The entirely of the mq/ include path is meant to be treated like library headers. Previously this implementation was inlined. Other parts of the headers are provided by mq2main, so it would make sense for this to be as well. If a plugin used these functions, where would it get its implementation from?

Knightly1: I removed `#include <Windows.h>` from this file so that I could use this file without bringing in Windows.h, but also so that I could use it elsewhere as a generic include.  The barrier for that are the two wstring functions.  Ideally those wouldn't be implemented in String.h at all, they'd be implemented in WString.h or something.  But since I don't know what (outside of our code base) is using those files, I opted to leave them there to be handled by implementation.  My preference would be to leave them inlined but move them to WString.h.

Our implementation of that relies on `#include <Windows.h>` which is in the WStringUtils.cpp file where the implementation of that is.  None of the other functions in String.h rely on anything outside of what is included (and all are std).  So this file is pretty portable as long as those functions aren't implemented using the windows lib (which for the majority of our uses they are).

Knightly1: > > since the implementation will differ on non-windows platforms.
> 
> I don't understand this statement at all XD

EQBCS for example compiles for Linux under gcc.

brainiac: the mq library headers aren't intended for linux targets, you need to provide your own implementation anyways, so i don't think it really applies here. 

I'm happy with how this is implemented now, it should just be exported so plugins have access to it too.

Knightly1: In this case, the majority of that file is portable (but I agree that the majority of MQ headers aren't nor should that be a design consideration) I was just explaining why I did it that way.  I will export it for plugin usage without the plugin requiring bringing in the implementation.

Knightly1: Or, I won't.  Re-review when you have a chance please.

brainiac (approved):

PR #702, opened 2023-03-09
Internal Path and loader for Lua

Sicprofundus:
- removed "2" from "openfolderMQ2" for launcher
- added lua to the open folder menu in the launcher
- added  Lua to internal paths

### DIFF
--- a/src/loader/MacroQuest.cpp
+++ b/src/loader/MacroQuest.cpp
@@ -212,6 +212,7 @@ std::string internal_paths::Logs = "Logs";
 std::string internal_paths::CrashDumps = internal_paths::Logs + "\\Dumps";
 std::string internal_paths::Plugins = "Plugins";
 std::string internal_paths::Resources = "Resources";
+std::string internal_paths::Lua = "Lua";
 
 bool InitializeDirectory(std::string& strPathToInit,
 	const std::string& iniKey, const std::string& iniFile,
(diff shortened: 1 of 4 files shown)


Knightly1:
This would be wrong for anyone who changes their lua path.

Sicprofundus:
> This would be wrong for anyone who changes their lua path. It's launching from the relative directory of the launcher and the path is never updated.

how do we correct that? Did a miss a part about how this already works for other folders?

brainiac:
Lua path is configurable in the lua plugin. It isn't like the other dirs.
https://github.com/macroquest/macroquest/blob/master/src/plugins/lua/MQ2Lua.cpp#L743-L765

Sicprofundus:
is the itent of "click the button to open your directory" still viable?

is this workable or should i close it?

Knightly1:
What brainiac said.  It's also not really relevant outside the context of the plugin, so the plugin provides the TLO for the path here: https://github.com/macroquest/macroquest/blob/master/src/plugins/lua/MQ2Lua.cpp#L345-L350

So you'd not have those paths in core.  To add them to the launcher, you'd probably want to export something from the plugin you can use to get the path.  Whether that is moving the config get to a shared header or actually exporting functionality I'd have to look at.

Sicprofundus:
closing MR

PR #716, opened 2023-05-01
Implement custom datatypes and TLOs in lua.

brainiac:
Also exposes `table` as a datatype from MQ2Lua so that other scripts/macros can access the data stored in lua tables exposed by custom datatypes.

Example custom TLO+datatype that implements a simple counter. There is another example in the pull request which includes a consumer that demonstrates cross-script communication.

```lua
local mq = require 'mq'

local counterType = mq.DataType.new('CounterDemoType', {
    Members = {
        Value = function(_, counter)
            return 'int', counter.value
        end,

        Name = function(_, counter)
            return 'string', counter.name
        end
    },

    Methods = {
        Reset = function(_, counter)
            counter.value = 0
        end,
        Increment = function(value, counter)
            counter.value = counter.value + tonumber(value)
        end
    }
})

function counterType.ToString(counter)
    return string.format('Counter %s: %d', counter.name, counter.value)
end

local allCounters = {
    default = { name = 'default', value = 0 }
}
local function CounterDemo(index)
    local name = index
    if index == nil or #index == 0 then
        name = 'default'
    end

    local counter = allCounters[name]
    if counter ~= nil then
        return counterType, counter
    else
        allCounters[name] = {name = name, value = 0}
        return counterType, allCounters[name]
    end
end

local function AllCounters(_)
    return 'table', allCounters
end

mq.AddTopLevelObject('CounterDemo', CounterDemo)
mq.AddTopLevelObject('AllCountersDemo', AllCounters)

while true do
    mq.delay(1000)
end
```

```
> /echo ${CounterDemo}
Counter default: 0
> /echo ${CounterDemo[test]}
Counter test: 0
> /invoke ${CounterDemo[test].Increment[3]}
> /echo ${CounterDemo[test]}
Counter test: 3
> /invoke ${CounterDemo[default].Increment[8]}
> /echo ${CounterDemo}
Counter default: 8
> /invoke ${CounterDemo.Reset}
> /echo ${CounterDemo}
Counter default: 0
> /echo ${AllCountersDemo}
table: 0x381405b0
> /echo ${AllCountersDemo.Keys}
table: 0x38149388
> /echo ${AllCountersDemo.Keys.Length}
2
> /echo ${AllCountersDemo.Keys.Index[1]}
test
> /echo ${AllCountersDemo.Keys.Index[2]}
default
> /echo ${AllCountersDemo.Key[test]}
table: 0x38143108
> /echo ${AllCountersDemo.Key[test].Key[name]}
test
> /echo ${AllCountersDemo.Key[test].Key[value]}
3.00
```

Known issues:

- [x] If script providing a TLO/datatype is killed while another script is consuming it, the client may crash


LemonStream:
Crash:
/echo ${CounterDemo.Increment} image

brainiac:
I can't do anything with this, can you post the call stack?

LemonStream:
Here you. 
 	```
KernelBase.dll!00007ffadabf4f69()	Unknown
>	MQ2Lua.dll!lj_err_throw(lua_State * L, int errcode) Line 470	C
 	MQ2Lua.dll!lj_err_run(lua_State * L) Line 569	C
 	MQ2Lua.dll!err_msgv(lua_State * L, ErrMsg em, ...) Line 582	C
 	MQ2Lua.dll!lj_err_optype(lua_State * L, const TValue * o, ErrMsg opm) Line 618	C
 	MQ2Lua.dll!lj_meta_arith(lua_State * L, TValue * ra, const TValue * rb, const TValue * rc, unsigned int op) Line 220	C
 	[External Code]	
 	[Inline Frame] MQ2Lua.dll!sol::basic_function<sol::basic_reference<0>,0>::luacall(__int64) Line 18948	C++
 	[Inline Frame] MQ2Lua.dll!sol::basic_function<sol::basic_reference<0>,0>::invoke(sol::types<>) Line 18971	C++
 	[Inline Frame] MQ2Lua.dll!sol::basic_function<sol::basic_reference<0>,0>::call(char * &) Line 19075	C++
 	[Inline Frame] MQ2Lua.dll!sol::basic_function<sol::basic_reference<0>,0>::operator()(char * &) Line 19061	C++
 	MQ2Lua.dll!mq::lua::bindings::LuaProxyType::GetMember(mq::MQVarPtr VarPtr, const char * Member, char * Index, mq::MQTypeVar & Dest) Line 743	C++
 	MQ2Main.dll!mq::EvaluateMacroDataMember(mq::datatypes::MQ2Type * type, mq::MQVarPtr & VarPtr, mq::MQTypeVar & Result, const std::string & Member, char * pIndex, bool checkFirst) Line 300	C++
 	MQ2Main.dll!mq::EvaluateDataExpression(mq::MQTypeVar & Result, const char * pStart, char * pIndex, bool function_allowed) Line 489	C++
 	MQ2Main.dll!ParseMQ2DataPortion(char * szOriginal, mq::MQTypeVar & Result) Line 625	C++
 	MQ2Main.dll!ParseMacroData(char * szOriginal, unsigned __int64 BufferSize) Line 1500	C++
 	MQ2Main.dll!ParseMacroParameter(eqlib::PlayerClient * pChar, char * szOriginal, unsigned __int64 BufferSize) Line 54	C++
 	[Inline Frame] MQ2Main.dll!mq::ParseMacroParameter(eqlib::PlayerClient *) Line 233	C++
 	MQ2Main.dll!mq::CCommandHook::Detour(eqlib::PlayerClient * pChar, const char * szFullLine) Line 408	C++
 	[External Code]	
 	MQ2Main.dll!mq::KeypressHandlerHook::HandleKeyDown_Hook(const eqlib::KeyCombo & combo) Line 169	C++
 	[External Code]	
 	MQ2Main.dll!mq::DoGameEventsPulse(int(*)() pEventFunc) Line 704	C++
 	MQ2Main.dll!mq::Detour_ProcessGameEvents() Line 741	C++
 	[External Code]	

```

File: src/plugins/lua/LuaImGui.cpp [hunk 1]
```diff
@@ -81,56 +81,19 @@ LuaImGui::~LuaImGui()
...
 	bool success = true;
 	try
 	{
-		int yield_count = 0;
-		constexpr int max_yields = 20;
 		ScopedYieldDisabler disableYield(LuaThread::get_from(m_thread.state()));
 
-		sol::protected_function_result result;
-		do
+		sol::function_result result = m_coroutine();
```
aquietone, line 91: This change from protected seems to result in lua stacktraces no longer printing on release builds, is that intentional? Still outputs fine on debug builds. On release, you just get the popup and message about how to resume the overlay but then no indication of the error that caused it.

brainiac: image
screenshot so you can see the highlighting.

brainiac: to translate, this change is a no-op

brainiac: ok yeah, looks like release builds change the SOL_SAFE_FUNCTION_OBJECTS_I_ define. thats not what i want..

src/plugins/lua/lua/examples/datatypes.lua
```diff
@@ -0,0 +1,123 @@
...
+    Methods = {
+        Echo = function (value) printf('Echo: %s', value) end,
+        ResetCounter = ResetCounter,
+    }
+})
+
+-- Additionally, the ToString function may be overridden to customize the stringified
+-- representation of your datatype
+function myType.ToString(value)
+    local result = '(Nothing)'
```
Knightly1, line 79: I realize this is an example of an override, where (Nothing) might be a valid return, but as a general example it's not a best practice I'd want to do without commenting it specifically.

src/plugins/lua/LuaImGui.cpp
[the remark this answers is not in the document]
brainiac: ScopedYieldDisabler disables the yielding, so this cod never actually yields.

src/plugins/lua/bindings/lua_MQMacroData.cpp
```diff
@@ -394,6 +600,435 @@ std::optional<std::string> mq_gettype_MQTypeVar(const lua_MQTypeVar& item)
...
+			}
+		}
+	}
+
+	return false;
+}
+
+bool lua_AddTopLevelObject(const char* name, sol::function func, sol::this_state s)
+{
+	auto& [_1, _2, numArgs, varargs] = GetArgInfo(func);
```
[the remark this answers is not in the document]
brainiac: well i can't use _ twice, so..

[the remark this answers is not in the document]
Knightly1: Just wondering if _1, _2 is more idiomatic than say "unused1, unused2" or "dummy1, dummy2" etc.  Your call either way, just curiosity given the normal nature of _variable name in MSVC.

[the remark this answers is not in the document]
brainiac: I copied this from dannuic, i don't have a preference.

include/mq/api/MacroAPI.h
```diff
@@ -140,6 +182,19 @@ inline namespace deprecated
...
 	}
+
+	//DEPRECATE("ParseMacroParameter no longer takes a PlayerClient parameter. It can be safely removed")
+	inline char* ParseMacroParameter(eqlib::PlayerClient* pChar, char* szOriginal, size_t BufferSize)
+	{
+		return mq::ParseMacroParameter(szOriginal, BufferSize);
+	}
+
+	template <unsigned int Size>
+	//DEPRECATE("ParseMacroParameter no longer takes a PlayerClient parameter.It can be safely removed")
```
Knightly1, line 193: I assume these two comments are for future, but what's the reason not to go ahead and deprecate them with this change?

include/mq/api/MacroDataTypes.h
```diff
@@ -0,0 +1,553 @@
...
+		}
+
+		// otherwise, just set the extra storage information and move on
+		HighPart_ = Val;
+		return HighPart_;
+	}
+
+#pragma warning(disable: 4996)
+	__declspec(property(get = get_HighPart, put = set_HighPart)) uint32_t HighPart;
+#pragma warning(default: 4996)
```
Knightly1, line 332: Why suppress above and disable/default here?  Leftover from the original copy?

src/main/MQ2ParseAPI.cpp
```diff
@@ -1,58 +0,0 @@
...
-{
-	DebugSpewNoFile("ShutdownParser()");
-
-	ShutdownMQ2Data();
-	ShutdownMQ2DataTypes();
-}
-
-char* ParseMacroParameter(PSPAWNINFO pChar, char* szOriginal, size_t BufferSize)
-{
-	EnterMQ2Benchmark(bmParseMacroParameter);
```
Knightly1, line 50: I don't personally care about this, but all of your changes derived from the ParseMacroParameter to ParseMacroData lose the EnterBenchMark code.  Just want to be sure that's intentional (I assume it is).

include/mq/base/String.h
```diff
@@ -690,4 +728,110 @@ inline bool GetBoolFromString(const std::string_view svString, const bool defaul
...
+	if (l != lsv.end()) return +1;
+	return 0;
+}
+
+/**
+   Functor class to compare two objects with the "Alphanum
+   Algorithm".
+*/
+template <class Ty>
+struct alphanum_less
```
Knightly1, line 829: Did I miss where these functions are used?

include/mq/api/MacroDataTypes.h
brainiac: guess so. will fix up

src/main/MQ2ParseAPI.cpp
brainiac: no, it shouldn't have been. it should've been the opposite (ParseMacroData didn't use the benchmark.) Will look it over again.

include/mq/base/String.h
brainiac: moved from one of the mq2xxx headers

dannuic (approved):

include/mq/api/MacroDataTypes.h
```diff
@@ -0,0 +1,552 @@
...
+	template <typename From, typename To>
+	static void ConvertData(const From& input, std::optional<To>& output)
+	{
+		if constexpr (std::is_convertible_v<From, To>)
+			output = static_cast<To>(input);
+		else
+		{
+			WriteChatf("Tried to convert unlike types %s and %s", type_name<From>().c_str(), type_name<To>().c_str());
+
+			if (gMacroBlock != nullptr && gMacroBlock->Line.find(gMacroBlock->CurrIndex) != gMacroBlock->Line.end())
```
dannuic, line 69: MQ2Globals.h needs to be included for this to make sense, should that just be included in this header?

dannuic: by "this" I mean using WriteChatf and gMacroStack

PR #719, opened 2023-05-09
Add a predicate search function for lua to find ground items

projecteon:
Functionality was mentioned and wished for by CB

https://discord.com/channels/511690098136580097/866047684242309140/1100025397346836482

### DIFF
--- a/src/main/datatypes/MQ2GroundType.cpp
+++ b/src/main/datatypes/MQ2GroundType.cpp
@@ -380,13 +380,8 @@ bool MQ2GroundType::dataItemTarget(const char* szIndex, MQTypeVar& Ret)
 MQTypeVar MQ2GroundType::MakeTypeVar(MQGroundSpawn groundSpawn)
 {
 	MQTypeVar Dest;
-
 	Dest.Type = pGroundType;
-	if (groundSpawn)
-		Dest.Set(groundSpawn);
-	else
-		Dest.Ptr = nullptr;
-
+	Dest.Set(groundSpawn);
 	return Dest;
 }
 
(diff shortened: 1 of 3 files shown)


src/main/datatypes/MQ2GroundType.cpp
```diff
@@ -377,4 +377,17 @@ bool MQ2GroundType::dataItemTarget(const char* szIndex, MQTypeVar& Ret)
...
 
+MQTypeVar MQ2GroundType::MakeTypeVar(MQGroundSpawn groundSpawn)
+{
+	MQTypeVar Dest;
+
+	Dest.Type = pGroundType;
+	if (groundSpawn)
+		Dest.Set(groundSpawn);
+	else
+		Dest.Ptr = nullptr;
```
brainiac, line 388: Should remove the condition. We want this to be a null ground spawn, not a nullptr  when it doesn't exist.

src/plugins/lua/bindings/lua_MQBindings.cpp
```diff
@@ -247,6 +247,28 @@ static sol::table lua_getFilteredSpawns(sol::this_state L, std::optional<sol::fu
 	return table;
 }
 
+static sol::table getFilteredGroundItems(sol::this_state L, std::optional<sol::function> predicate)
```
Knightly1, line 250: Let's call this lua_getFilteredGroundItems for consistency.

Knightly1:
For parity, I've added Issue #728 .  This PR is fine without it, but it's something that should be addressed to maintain parity.

PR #720, opened 2023-05-09
Has Skill return value

Sicprofundus:
- corrected inline bool return value from "0" to false

### DIFF
--- a/src/main/MQ2Inlines.h
+++ b/src/main/MQ2Inlines.h
@@ -249,7 +249,7 @@ inline int GetFocusRangeModifier(const EQ_Spell* pSpell, ItemPtr& pItemOut)
 
 inline bool HasSkill(int nSkill)
 {
-	return pLocalPC ? pLocalPC->HasSkill(nSkill) : 0;
+	return pLocalPC ? pLocalPC->HasSkill(nSkill) : false;
 }
 
 inline float GetDistance(float X1, float Y1)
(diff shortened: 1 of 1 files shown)


Sicprofundus:
@brainiac sorry for deleting and recommitting, something happened with my git this morning i had to fix.

no clue where the asd came from, but i appreciate your eyeballs

src/main/MQ2Inlines.h
```diff
@@ -249,7 +249,7 @@ inline int GetFocusRangeModifier(const EQ_Spell* pSpell, ItemPtr& pItemOut)
 
 inline bool HasSkill(int nSkill)
 {
-	return pLocalPC ? pLocalPC->HasSkill(nSkill) : 0;
+	return pLocalPC ? pLocalPC->HasSkill(nSkill) : false;
```
brainiac, line 252: this can be written without the ternary `return pLocalPC && pLocalPC->HasSkill(nSkill)`

Sicprofundus: I appreciate you pointing that out, sometimes i have a bad habit of not stepping back for the bigger "what are we trying to do" and instead see a small part of it

dannuic: what _are_ you trying to do? This isn't a functional change

Sicprofundus: it was returning a  "0" for a bool function, so i was trying to clean it up to return false instead of 0

PR #724, opened 2023-06-03
Update MQ2MathType.cpp with comma and space for distance separation

bytebite01:
Changed singular commas into comma or spaces and eats leading comma or spaces.

### DIFF
--- a/src/main/datatypes/MQ2MathType.cpp
+++ b/src/main/datatypes/MQ2MathType.cpp
@@ -248,51 +248,51 @@ bool MQ2MathType::GetMember(MQVarPtr VarPtr, const char* Member, char* Index, MQ
 			if (char* pColon = strchr(Index, ':'))
 			{
 				*pColon = 0;
-				if (char* pComma = strchr(&pColon[1], ','))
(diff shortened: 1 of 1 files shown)


brainiac (changes requested):
At this rate, we should just factor out the code to parse a coordinate from a string, then just call it once or twice for however many coordinate we expect there to be. 

Then we'd simply need to split the string, then parse a point from one or both halves. Parsing the point can then split on commas or empty space. should be pretty simple and minimalize the amount of repetition.

brainiac:
bonus points for implementing this using string_view instead of raw char *

src/main/datatypes/MQ2MathType.cpp
```diff
@@ -14,6 +14,7 @@
 
 #include "pch.h"
 #include "MQ2DataTypes.h"
+#include "string_view"
```
brainiac, line 17: standard library headers should be included with <angle brackets>

brainiac:
This PR was actually fine until your most recent change. You can't mutate the input string. So unfortunately I can't accept this as is. If you revert that last commit it should be ok.

brainiac (approved):

Issue #96, opened 2020-04-10
Calling a macro from another macro misses the first line


MacroQuestGitHubMigrationBot:
In GitLab by @Knightly1 on Apr 10, 2020, 09:32

Confirm if this is a bug in Next as well (it should be, that portion of the macro engine hasn't changed).

PR #725, opened 2023-06-11
Suggested fix #703 endmac from sub returning to wrong spot.

### LINKED ISSUE #703: Macro - /endmac called from a Sub returns OnExit to the wrong place
**Describe the bug**
When /endmac is called from a Sub and :OnExit exists, the sub returns to its exit position and continues to execute.  Instead, it should only execute the OnExit code and not continue from the Sub return.

**To Reproduce**
```
-- issue700.mac
#include issue700.inc
Sub Main()
    /call TestEnd
    /echo Test Failed
:OnExit
   /call MainExit
/return

Sub TestEnd()
    /echo Testing endmac
    /endmac
/return

Sub MainExit()
   /echo In MainExit
/return
````

`/mac issue700` output:
```
Testing endmac
Test Failed
In MainExit
The current macro has ended.
```

**Expected behavior**
The expected output would be to stop executing the main sub and jump directly to OnExit resulting in:
```
Testing endmac
In MainExit
The current macro has ended.
```

**MQ Version**
- [X] Live
- [X] Test
- [X] Emu (ROF2)

### PR DESCRIPTION
**Pull request by bytebite01**
Removed increment to next line after :OnExit since MQ2Pulse increments after Call command and a Sub calling /endmacro would result in losing 1 line of macro.

:OnExit must end with an /endmacro line since the presence of :OnExit catches the first call to /endmacro and halts it after unpausing and setting current index to :OnExit's location.

### DIFF
--- a/src/main/MQ2MacroCommands.cpp
+++ b/src/main/MQ2MacroCommands.cpp
@@ -1346,7 +1346,6 @@ void EndMacro(PSPAWNINFO pChar, char* szLine)
 	{
 		if (!_stricmp(":OnExit", i->second.Command.c_str()))
 		{
-			i++;
 			pBlock->CurrIndex = i->first;
 			// Force unpause to finish processing
 			pBlock->Paused = false;

### CODE REVIEW

**Comment by Knightly1** [comment]
While I agree this resolves an issue, this doesn't resolve Issue #703.  I believe this would change the current output of Issue #703 from:

```
Testing endmac
Test Failed
In MainExit
The current macro has ended.
```

To:

```
Testing endmac
In MainExit
Test Failed
In MainExit
The current macro has ended.
```

Resulting in MainExit being called twice, unless your /endmac is added to end the macro a second time.  The resolution for Issue #703 requires the output to be:

```
Testing endmac
In MainExit
The current macro has ended.
```

without changing the code as outlined in the issue.  Adding the additional requirement of having :OnExit end with /endmacro is just obfuscating the issue.

Good research so far, but this would introduce a new bug while not really fixing the old one.  It's also possible the issue you're looking at now is related to #96 (though I didn't look deep enough to see).

I like where you're going but accepting this Pull Request would be a breaking change.  I'm going to go ahead and create another issue for the other issue you found though and let me know if I've misread or misunderstood your change.

Knightly1:
I've updated this to reference #726 - but, as I mentioned, it becomes a breaking change without fixing the issue in #703

bytebite01:
#726 comment includes test cases.  This should fix #703 but I'm feeling like a 3 year old messing with parents' tools.

bytebite01:
Fixed the stack cleanup to keep current context instead of return to Main context on /endMacro when :OnExit is present to match with current behavior.  #703 and #726 fixes should pass those tests in #726.  

The only concern I have remaining is that /endmacro from the console still executes the label line.  Mostly a non issue but might be failing enough to not warrant a change.

Knightly1:
> The only concern I have remaining is that /endmacro from the console still executes the label line. Mostly a non issue but might be failing enough to not warrant a change.

I've been planning to look at this.  When you say "executes" is this a user facing issue or just something you noticed in debugging?

bytebite01:
Completely transparent to user.  Just debugging.

Knightly1:
Added some comments on #726 so that it's in context with your other tests.

PR #727, opened 2023-06-15
Normalize GetDistance as the method for distance

Knightly1:
- Consolidate and deprecate other forms of GetDistance
- GetDistance should be used for coordinates and overloads handle 2D/3D
- GetDistanceFromPlayer is added to be more explicit since it is not comparing distance between two coordinates, but rather to the player
- GetDistance2D and GetDistance3D are explicit when overloads cannot discern intent
- Add Documentation for each Distance related function
- Ensure every GetDistance form has a GetDistanceSquared counterpart


brainiac:
We should move these distance functions out into something like mq/base/Math.h

src/common/StringUtils.h
```diff
@@ -14,18 +14,77 @@
 
 #pragma once
 
+// The functions that are here, and not in mq/base/String.h depend on third
+// party libraries.
```
brainiac, line 18: Not sure if this was intentional, but src/common/StringUtils.h by convention is a private header file

Knightly1:
> We should move these distance functions out into something like mq/base/Math.h

Agreed.  I considered doing it with this, but I wanted some eyes on the code too to make sure everything was headed down the right path.  I'll make that change.

src/main/MQ2Inlines.h
```diff
@@ -252,90 +252,394 @@ inline bool HasSkill(int nSkill)
...
+ * @param first The first glm::vec3 point.
+ * @param second The second glm::vec3 point.
+ * @return float The 3D Euclidean distance between the two glm::vec3 points.
+ */
+inline float GetDistance(const glm::vec3& first, const glm::vec3& second)
+{
+	return GetDistance(first.x, first.y, first.z, second.x, second.y, second.z);
+}
+
+[[deprecated("Use GetDistance instead.")]]
```
brainiac, line 311: Use the DEPRECATE macro, for some reason these may cause issues with the clang offset generator

src/common/StringUtils.h
Knightly1: Unintentional.  I'll evaluate.

src/main/MQ2Data.cpp
```diff
@@ -148,75 +148,31 @@ bool dataSubDefined(const char* szIndex, MQTypeVar& Ret)
...
-					DestPos.Y = GetFloatFromString(&pComma[1], DestPos.Y);
-				}
-			}
-			else
-			{
-				DestPos.X = GetFloatFromString(&pColon[1], DestPos.X);
-			}
-		}
+			const glm::vec3 first_loc = GetVec3FromString(args[0], { pControlledPlayer->Y, pControlledPlayer->X, pControlledPlayer->Z });
+			glm::vec3 second_loc = { 0.0f, 0.0f, 0.0f };
```
bytebite01, line 163: Old behavior if no second location is supplied defaulted to Player X Y Z.  new one seems to compare against 0,0,0 ?

src/main/datatypes/MQ2MathType.cpp
```diff
@@ -238,29 +238,23 @@ bool MQ2MathType::GetMember(MQVarPtr VarPtr, const char* Member, char* Index, MQ
...
-			P[0][1] = P[1][1] = pControlledPlayer->X;
-			P[0][2] = P[1][2] = pControlledPlayer->Z;
-
-			for (int i = 0; i < p_list.size() && i < 2; i++)				// for ever separate location
+			if (!args.empty())
 			{
-				auto pointList = split_view(p_list[i], ' ', true);			// create a string view list of each float
-				for (int j = 0; j < pointList.size() && j < 3; j++)			// for every string broken by spaces
+				const glm::vec3 first_loc = GetVec3FromString(args[0]);
+				glm::vec3 second_loc = { 0.0f, 0.0f, 0.0f };
```
bytebite01, line 248: If no second arg is supplied, default was Player Y , X, Z.  Change makes it 0, 0, 0

Knightly1: Good catch, my bad.  I had that in my tests but I was testing from 0,0,0.

src/common/StringUtils.h
Knightly1: Instead of putting these here, I put them in mq/libutils/glmUtils.h

src/main/MQ2Inlines.h
Knightly1: Done

src/main/MQ2Data.cpp
Knightly1: Good catch, but to confirm it was Y, X, Z, before yeah?  Can you check current changes and make sure I didn't mess that up?

src/main/datatypes/MQ2MathType.cpp
Knightly1: Should be resolved in latest.

src/main/MQ2Data.cpp
bytebite01: Confirmed.  Way it was done before was confusing with Y value in variable X.

src/main/MQ2Data.cpp
```diff
@@ -148,73 +149,21 @@ bool dataSubDefined(const char* szIndex, MQTypeVar& Ret)
...
-			{
-				SourcePos.Y = GetFloatFromString(&pComma[1], SourcePos.Y);
-			}
-		}
-		else
-		{
-			SourcePos.X = GetFloatFromString(szTemp, SourcePos.X);
-		}
+		const glm::vec3 player_loc = { pLocalPlayer->Y, pLocalPlayer->X, pLocalPlayer->Z };
+		const auto coordinates = GetVec3SetFromString(szIndex, player_loc, player_loc);
```
brainiac, line 161: why not use structured bindings here?

`const auto [coord1, coord2] = ...`

src/main/MQ2Data.cpp
Knightly1: Thanks, just wanted to be sure.  I agree it was confusing before.  That's actually why I opted to use the coord1[0] notation instead of coord1.x (because the swap itself is already confusing)

src/main/MQ2Data.cpp
Knightly1: You and your C++ 17 making things more readable.

Done.

Knightly1:
Resolved merge conflicts and brought PR up to date. Updated the include/mq/README.md with current state.

include/mq/base/Math.h
```diff
@@ -0,0 +1,407 @@
...
+	return GetDistance(first.x, first.y, first.z, second.x, second.y, second.z);
+}
+
+DEPRECATE("Use GetDistance instead.")
+inline float GetDistance3D(float X1, float Y1, float Z1, float X2, float Y2, float Z2)
+{
+	return GetDistance(X1, Y1, Z1, X2, Y2, Z2);
+}
+
+DEPRECATE("Use GetDistance instead.")
```
brainiac, line 70: this macro requires an include

include/mq/base/Math.h
```diff
@@ -0,0 +1,407 @@
...
+ * @brief Calculates and returns the 2D Euclidean distance between the positions of two spawns.
+ *
+ * @note If you don't need the actual distance, but only need to compare distances, you should
+ * use GetDistanceSquared2D instead.
+ *
+ * @param spawn1 The first spawn.
+ * @param spawn2 The second spawn.
+ * @return float The 2D Euclidean distance between the positions of the two spawns.
+ */
+inline float GetDistance2D(const SPAWNINFO* spawn1, const SPAWNINFO* spawn2)
```
brainiac: Use the non-typedef, include the necessary header.

include/mq/base/Math.h
```diff
@@ -0,0 +1,407 @@
...
+ * @brief [3D] Calculates and returns the squared Euclidean distance between two glm::vec3 points in a 3D space.
+ *
+ * @note This function is often used when comparing distances, as squaring is cheaper than taking the square root.
+ * If you actually need the distance, use the GetDistance function instead.
+ *
+ * @param first The glm::vec3 coordinate of the first point.
+ * @param second The glm::vec3 coordinate of the second point.
+ * @return float The squared Euclidean distance between the two points.
+ */
+inline float GetDistanceSquared(const glm::vec3& first, const glm::vec3& second)
```
brainiac: glm needs a header include

include/mq/base/Math.h
Knightly1: Deprecate the typedef so when I copy 3 year old code verbatim it says "Hey, maybe look at your code." 😆 (Fixed)

include/mq/base/Math.h
Knightly1: Done

include/mq/base/Math.h
Knightly1: Done.

include/mq/base/Math.h
brainiac: i've thought about it but so much code would need to be changed i'm not sure its worth. probably should do a cleanup pass on MQ first before resorting to deprecations.

brainiac:
it only took three years but we got there

PR #733, opened 2023-07-14
Added FG to AdvLootType

PhantomRogue:
Updated to add FG attribute to the AdvLootType.

Can check if an Item is set to FG in the Personal/Shared Lists.

### DIFF
--- a/src/main/datatypes/MQ2AdvLootType.cpp
+++ b/src/main/datatypes/MQ2AdvLootType.cpp
@@ -212,7 +212,7 @@ enum class AdvLootItemMembers
 	Never,
 	IconID,
 	NoDrop,
-	FG
+	FreeGrab
 };
 
 MQ2AdvLootItemType::MQ2AdvLootItemType() : MQ2Type("advlootitem")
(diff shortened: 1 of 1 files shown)


src/main/datatypes/MQ2AdvLootType.cpp
```diff
@@ -211,7 +211,8 @@ enum class AdvLootItemMembers
 	AlwaysGreed,
 	Never,
 	IconID,
-	NoDrop
+	NoDrop,
+	FG
```
brainiac, line 215: what does FG mean in this context?

Sicprofundus: Free Grab (should call it that)

brainiac: Yes lets call it that.

brainiac (changes requested):
Lets call this FreeGrab instead of FG so that its clear what this means

src/main/datatypes/MQ2AdvLootType.cpp
dannuic: Might want to change it on the item as well

dannuic: I was just informed this was direct eqlib data, not an MQ construct, so changing the name upstream is probably out of scope.

Sicprofundus: > I was just informed this was direct eqlib data, not an MQ construct, so changing the name upstream is probably out of scope.

what's best to do here? overload it so FG and FreeGrab exist? or just leave it FG?

dannuic: Just change the datatype member to FreeGrab

brainiac: dannuic was talking about eqlib which is the source data and is out of scope for this pull request. The request to change FreeGrab covers the changes that are already in this pull request.

PhantomRogue:
Ok, will change the Name to FreeGrab instead of FG.

brainiac (approved):

PR #737, opened 2023-07-19
Expose profile data for use in TLO AutoLogin, including hotkey.

htwmmo:
Originally I was only exposing the hotkey, and saving it in the TLO class to a char array, and the TLO had only member HotKey.  The intent was to use it for things like /setwintitle, to see server, char, zone, hotkey, etc. to make clear while getting used to boxing groups, which char was on which hotkey.  Decided to just expose the additional data (other than password) via the TLO anyway.  The only imgui addition to the display for autologin I added was hotkey.

### DIFF
--- a/src/plugins/autologin/MQ2AutoLogin.cpp
+++ b/src/plugins/autologin/MQ2AutoLogin.cpp
@@ -45,12 +45,13 @@ uint64_t ReenableTime = 0;
 int Level = -1;
 int Class = -1;
 
+std::optional<ProfileRecord> m_record_tlo = std::nullopt;
+
 class MQ2AutoLoginType* pAutoLoginType = nullptr;
 
 class MQ2AutoLoginType : public MQ2Type
 {
 public:
-	std::optional<ProfileRecord> m_record;
 	enum class AutoLoginMembers
 	{
 		HotKey,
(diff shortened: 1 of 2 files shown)


src/plugins/autologin/MQ2AutoLogin.cpp
```diff
@@ -45,6 +45,96 @@ uint64_t ReenableTime = 0;
 int Level = -1;
 int Class = -1;
 
+class MQ2AutoLoginType* pAutoLoginType = nullptr;
+
+class MQ2AutoLoginType : public MQ2Type
+{
+public:
+	std::optional<ProfileRecord> m_record;
```
brainiac, line 53: datatype's should not hold state. this should be made a global variable instead, or it should be passed through the VarPtr

src/plugins/autologin/MQ2AutoLogin.cpp
```diff
@@ -617,6 +712,7 @@ PLUGIN_API void OnPulse()
 			{ "news",                 "NEWS_OKButton"}
 		};
 
+		memcpy_s(&pAutoLoginType->m_record, sizeof(std::optional<ProfileRecord>), (const void *)Login::getrecord(), sizeof(std::optional<ProfileRecord>));
```
brainiac, line 715: This isn't legal -- you can't memcpy a non-pod type. You should be able to just perform a straight assignment here.

Is this meant to capture the record that we're using to log in with? It might be better to assign this where we set the profile with SetLoginInformation

brainiac (changes requested):
I like the concept, there are a couple small issues though. see comments

src/plugins/autologin/MQ2AutoLogin.cpp
```diff
@@ -45,6 +45,96 @@ uint64_t ReenableTime = 0;
...
+	virtual bool GetMember(MQVarPtr VarPtr, const char* Member, char* Index, MQTypeVar& Dest) override
+	{
+		MQTypeMember* pMember = MQ2AutoLoginType::FindMember(Member);
+		if (!pMember)
+			return false;
+
+		switch ((AutoLoginMembers)pMember->ID)
+		{
+		case AutoLoginMembers::HotKey:
+			strcpy_s(DataTypeTemp, m_record->hotkey.data());
```
brainiac, line 85: Oh, almost missed this -- since m_record is an `optional`, it could be unset. You should check for that and bail at the top of GetMember

src/plugins/autologin/MQ2AutoLogin.cpp
htwmmo: Moved to global var.

src/plugins/autologin/MQ2AutoLogin.cpp
htwmmo: Changed to assign directly via getter.

src/plugins/autologin/MQ2AutoLogin.cpp
htwmmo: Added check before switch.

htwmmo (commented):
Made changes per review, see commit comments.

brainiac (approved):

brainiac:
I moved the profile properties into a subtype, so they are accessed via something like AutoLogin.Profile.Character. Not 100% sold on the name "Profile" so I might change it again before cutting a release...

PR #753, opened 2023-09-16
Unpickle

Atomfusion1:
We Got Pickle to save file.. 
this gives unpickle to load the file back
one line, No trial error to find out why its not loading when it points to EQ instead of eq/config 

2 changes
1. minor changed pickle from void to return bool, return true if load, else false
2. lua_unpickle . getconfig path, merge path if fullname,  convert to double \\ . run lua loadfile

### DIFF
--- a/src/plugins/lua/bindings/lua_MQBindings.cpp
+++ b/src/plugins/lua/bindings/lua_MQBindings.cpp
@@ -515,7 +515,7 @@ static void serialize(sol::object obj, int prefix_count, fmt::appender& appender
 	}
 }
 
-static void lua_pickle(sol::this_state L, std::string_view file_path, sol::table table)
+static bool lua_pickle(sol::this_state L, std::string_view file_path, sol::table table)
 {
 	fmt::memory_buffer buf;
 	fmt::appender appender(buf);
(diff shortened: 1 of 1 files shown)


src/plugins/lua/bindings/lua_MQBindings.cpp
```diff
@@ -529,21 +529,42 @@ static void lua_pickle(sol::this_state L, std::string_view file_path, sol::table
...
+	for (size_t pos = 0; (pos = pathString.find("\\", pos)) != std::string::npos; pos += 2) 
+	{
+		pathString.replace(pos, 1, "\\\\");
+	}
+
+	// Concatenate strings and run Lua code:
+	std::string luaCode =
+		"local fn, err = loadfile('" + pathString + "')\n"
+		"if not fn then error(err or '') end\n"
+		"return fn()";
```
brainiac, line 564: If you went this far you might as well just call this on the c++ side. All you're doing here is writing a script that does nothing but call one function, and we can do that better from c++.

brainiac (changes requested):
see comment

dannuic:
At this point, we should probably have a whole pickle namespace to provide all the pickle related functions, dump and load

brainiac:
i don't know if thats necessary...

PR #757, opened 2023-10-01
Add new Bank TLO

projecteon:
I've left shard bank intentionally as I thought it should be its own TLO.


src/main/datatypes/MQ2BankType.cpp
```diff
@@ -0,0 +1,160 @@
...
+	switch (static_cast<BankTypeMembers>(pMember->ID))
+	{
+	case BankTypeMembers::FreeSlots:
+	{
+		int nSize = 0;
+		if (Index[0])
+		{
+			nSize = GetIntFromString(Index, 0);
+			if (nSize > 4)
+				nSize = 4;
```
brainiac, line 90: this should be a constant from the eItemSize enumeration

src/main/datatypes/MQ2BankType.cpp
```diff
@@ -0,0 +1,160 @@
...
+
+			if (nSlot < GetAvailableBankSlots())
+			{
+				Ret = pItemType->MakeTypeVar(pLocalPC->BankItems.GetItem(nSlot));
+				return true;
+			}
+		}
+	}
+
+	// If no index is provided then we're utilizing the achievement manager type
```
brainiac, line 155: this and the previous comment refer to achievments.

src/main/datatypes/MQ2BankType.cpp
```diff
@@ -0,0 +1,160 @@
...
+	Copper,
+};
+
+int GetNumFreeBankSlots(int nSize)
+{
+	if (!pLocalPC || !pLocalPlayer)
+		return 0;
+
+	int freeSlots = 0;
+	for (int slot = 0; slot < GetAvailableBankSlots(); slot++)
```
Knightly1, line 39: GetAvailableBankSlots() isn't super intensive or anything, but it's also not going to change:

```cpp
int availableBankSlots = GetAvailableBankSlots();
for (int slot = 0; slot < availableBankSlots; ++slot)
```

src/main/datatypes/MQ2BankType.cpp
```diff
@@ -0,0 +1,160 @@
...
+
+			if (nSlot < GetAvailableBankSlots())
+			{
+				Ret = pItemType->MakeTypeVar(pLocalPC->BankItems.GetItem(nSlot));
+				return true;
+			}
+		}
+	}
+
+	// If no index is provided then we're utilizing the bank type
```
Knightly1, line 155: The logic here doesn't match the comment -- this actually occurs if there's no index, or if the index is not a number, or if the number on index (minus 1) is greater than the available bank slots.  If you want this to be true, you probably need a return higher.

PR #760, opened 2023-10-26
Update MQ2GroupType.cpp

JBzone:
Expanded on the idea of Group.Cleric and added the remaining classes. Also added options Furthest, FurthestMelee, and FurthestPureCaster will return the 3D distance of the furthest player in the group of those types. To make it easier to figure out if group members are close enough for group type spells/effects. IE 2.0 clicks / Group heals.

### DIFF
--- a/src/main/datatypes/MQ2AdvLootType.cpp
+++ b/src/main/datatypes/MQ2AdvLootType.cpp
@@ -351,35 +351,41 @@ bool MQ2AdvLootItemType::GetMember(MQVarPtr VarPtr, const char* Member, char* In
 	{
 		Dest.Type = pStringType;
 
+		std::string tmp_str = "INVALID";
+
+		// Check to make sure we aren't trying to access the personal loot window and if so return INVALID.
 		if( itemList == pAdvancedLootWnd->pPLootList )
 		{
-			Dest.Ptr = "INVALID";
+			strcpy_s(DataTypeTemp, tmp_str.c_str());
+			Dest.Ptr = &DataTypeTemp[0];
 			return true;
 		}
+
 		switch( item.Unknown0x64 )
 		{
 		case 0:
-			Dest.Ptr = "WAITING";
-			return true;
+			tmp_str = "WAITING";
+			break;
 		case 1:
-			Dest.Ptr = "ASKING";
-			return true;
+			tmp_str = "ASKING";
+			break;
 		case 2:
-			Dest.Ptr = "ROLLING";
-			return true;
+			tmp_str = "ROLLING";
+			break;
 		case 3:
-			Dest.Ptr = "STOPPED";
-			return true;
+			tmp_str = "STOPPED";
+			break;
 		case 4:
-			Dest.Ptr = "CLICKROLL";
-			return true;
+			tmp_str = "CLICKROLL";
+			break;
 		case 5:
-			Dest.Ptr = "FREEGRAB";
-			return true;
-		default:
-			Dest.Ptr = "UNKNOWN";
-			return true;
+			tmp_str = "FREEGRAB";
+			break;
 		}
+
+		strcpy_s(DataTypeTemp, tmp_str.c_str());
+		Dest.Ptr = &DataTypeTemp[0];
+		return true;
 	}
 
 	default: 
--- a/src/main/datatypes/MQ2AdvLootType.cpp
+++ b/src/main/datatypes/MQ2AdvLootType.cpp
@@ -212,7 +212,8 @@ enum class AdvLootItemMembers
 	Never,
 	IconID,
 	NoDrop,
-	FreeGrab
+	FreeGrab,
+	Status
 };
 
 MQ2AdvLootItemType::MQ2AdvLootItemType() : MQ2Type("advlootitem")
@@ -232,6 +233,7 @@ MQ2AdvLootItemType::MQ2AdvLootItemType() : MQ2Type("advlootitem")
 	ScopedTypeMember(AdvLootItemMembers, IconID);
 	ScopedTypeMember(AdvLootItemMembers, NoDrop);
 	ScopedTypeMember(AdvLootItemMembers, FreeGrab);
+	ScopedTypeMember( AdvLootItemMembers, Status );
 }
 
 bool MQ2AdvLootItemType::GetMember(MQVarPtr VarPtr, const char* Member, char* Index, MQTypeVar& Dest)
@@ -345,7 +347,43 @@ bool MQ2AdvLootItemType::GetMember(MQVarPtr VarPtr, const char* Member, char* In
 		Dest.Type = pBoolType;
 		return true;
 
-	default: break;
+	case AdvLootItemMembers::Status:
+	{
+		Dest.Type = pStringType;
+
+		if( itemList == pAdvancedLootWnd->pPLootList )
+		{
+			Dest.Ptr = "INVALID";
+			return true;
+		}
+		switch( item.Unknown0x64 )
+		{
+		case 0:
+			Dest.Ptr = "WAITING";
+			return true;
+		case 1:
+			Dest.Ptr = "ASKING";
+			return true;
+		case 2:
+			Dest.Ptr = "ROLLING";
+			return true;
+		case 3:
+			Dest.Ptr = "STOPPED";
+			return true;
+		case 4:
+			Dest.Ptr = "CLICKROLL";
+			return true;
+		case 5:
+			Dest.Ptr = "FREEGRAB";
+			return true;
+		default:
+			Dest.Ptr = "UNKNOWN";
+			return true;
+		}
+	}
+
+	default: 
+		break;
 	}
 
 	return false;
--- a/src/main/datatypes/MQ2GroupType.cpp
+++ b/src/main/datatypes/MQ2GroupType.cpp
@@ -15,1125 +15,1121 @@
 #include "pch.h"
 #include "MQ2DataTypes.h"
 
-namespace mq::datatypes
+namespace mq::datatypes {
+
+enum class GroupMembers
 {
+	Member = 1,
+	Members,
+	Leader,
+	GroupSize,
+	MainTank,
+	MainAssist,
+	Puller,
+	MarkNpc,
+	MasterLooter,
+	AnyoneMissing,
+	Present,
+	MercenaryCount,
+	TankMercCount,
+	HealerMercCount,
+	MeleeMercCount,
+	CasterMercCount,
+	MouseOver,
+	AvgHPs,
+	Injured,
+	LowMana,
+	Cleric,
+	Shaman,
+	Warrior,
+	Paladin,
+	Ranger,
+	Shadowknight,
+	Druid,
+	Monk,
+	Bard,
+	Rogue,
+	Necromancer,
+	Wizard,
+	Mage,
+	Magician,
+	Enchanter,
+	Beastlord,
+	Berserker,
+	Furthest,
+	FurthestMelee,
+	FurthestPureCaster
+};
 
-	enum class GroupMembers
+MQ2GroupType::MQ2GroupType() : MQ2Type("group")
+{
+	ScopedTypeMember(GroupMembers, Member);
+	ScopedTypeMember(GroupMembers, Members);
+	ScopedTypeMember(GroupMembers, Leader);
+	ScopedTypeMember(GroupMembers, GroupSize);
+	ScopedTypeMember(GroupMembers, MainTank);
+	ScopedTypeMember(GroupMembers, MainAssist);
+	ScopedTypeMember(GroupMembers, Puller);
+	ScopedTypeMember(GroupMembers, MarkNpc);
+	ScopedTypeMember(GroupMembers, MasterLooter);
+	ScopedTypeMember(GroupMembers, AnyoneMissing);
+	ScopedTypeMember(GroupMembers, Present);
+	ScopedTypeMember(GroupMembers, MercenaryCount);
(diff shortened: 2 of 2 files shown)


src/main/datatypes/MQ2GroupType.cpp
```diff
@@ -15,768 +15,1125 @@
...
-	return true;
-}
+	}
 
-bool MQ2GroupType::GetMember(MQVarPtr VarPtr, const char* Member, char* Index, MQTypeVar& Dest)
-{
-	MQTypeMember* pMember = MQ2GroupType::FindMember(Member);
+	bool MQ2GroupType::ToString( MQVarPtr VarPtr, char* Destination )
+	{
+		if( !pLocalPC || !pLocalPC->Group ) return false;
```
Sicprofundus, line 112: what is with all the extra spaces around brakcets

dannuic: It looks like this got run through a formatter, which is making it hard to review (and makes the style inconsistent with the rest of the project). Please limit changes to actual code changes

src/main/datatypes/MQ2AdvLootType.cpp
```diff
@@ -345,7 +347,43 @@ bool MQ2AdvLootItemType::GetMember(MQVarPtr VarPtr, const char* Member, char* In
...
+		case 4:
+			Dest.Ptr = "CLICKROLL";
+			return true;
+		case 5:
+			Dest.Ptr = "FREEGRAB";
+			return true;
+		default:
+			Dest.Ptr = "UNKNOWN";
+			return true;
+		}
```
dannuic, line 382: You can't set pointers to string literals, you need to copy into a pointer and then set it (we use DataTypeTemp). Look at CharacterType.CombatState for an example.

src/main/datatypes/MQ2AdvLootType.cpp
```diff
@@ -345,7 +347,43 @@ bool MQ2AdvLootItemType::GetMember(MQVarPtr VarPtr, const char* Member, char* In
...
+	case AdvLootItemMembers::Status:
+	{
+		Dest.Type = pStringType;
+
+		if( itemList == pAdvancedLootWnd->pPLootList )
+		{
+			Dest.Ptr = "INVALID";
+			return true;
+		}
+		switch( item.Unknown0x64 )
```
dannuic, line 359: should this be renamed in eqlib?

dannuic:
Please undo the formatting changes in grouptype before I can review this

src/main/datatypes/MQ2GroupType.cpp
JBzone: ya sorry my VS settings copy / paste auto updates formatting. I can reform it.

src/main/datatypes/MQ2AdvLootType.cpp
brainiac: Yes. This cannot be used here as the name will change (on purpose - unknown values should not be referenced)

brainiac (changes requested):
The formatting and code errors need to be addressed. 

The group changes should be split apart from the autoloot changes so they can be reviewed separately.

src/main/datatypes/MQ2AdvLootType.cpp
```diff
@@ -345,7 +347,49 @@ bool MQ2AdvLootItemType::GetMember(MQVarPtr VarPtr, const char* Member, char* In
 		Dest.Type = pBoolType;
 		return true;
 
-	default: break;
+	case AdvLootItemMembers::Status:
+	{
+		Dest.Type = pStringType;
+
+		std::string tmp_str = "INVALID";
```
dannuic, line 354: Just copy the literal in the switch, there's no reason to allocate a temporary string

src/main/datatypes/MQ2AdvLootType.cpp
```diff
@@ -345,7 +347,49 @@ bool MQ2AdvLootItemType::GetMember(MQVarPtr VarPtr, const char* Member, char* In
...
+
+		// Check to make sure we aren't trying to access the personal loot window and if so return INVALID.
+		if( itemList == pAdvancedLootWnd->pPLootList )
```
dannuic, line 357: Is there a reason you are formatting like this? This is inconsistent with the rest of the code base -- it should be `if (thing)`

src/main/datatypes/MQ2AdvLootType.cpp
dannuic: Yeah, I don't think this could be merged until this is added to eqlib because of this

src/main/datatypes/MQ2GroupType.cpp
```diff
@@ -97,6 +135,8 @@ bool MQ2GroupType::GetMember(MQVarPtr VarPtr, const char* Member, char* Index, M
 		return false;
 	}
 
+	float furthest_dist = 0;
```
dannuic, line 138: This should be in the scope of the members that need it, there's no reason to declare it out here.

src/main/datatypes/MQ2GroupType.cpp
```diff
@@ -414,6 +454,261 @@ bool MQ2GroupType::GetMember(MQVarPtr VarPtr, const char* Member, char* Index, M
...
+			   && member->pSpawn
+			   && member->pSpawn->GetClass() == Shaman)
+			{
+				Dest = pSpawnType->MakeTypeVar(member->pSpawn);
+				return true;
+			}
+		}
+
+		Dest = pSpawnType->MakeTypeVar();
+		return true;
```
dannuic, line 473: There's a ton of repeated code in here, I'd turn this into a function

src/main/datatypes/MQ2GroupType.cpp
```diff
@@ -414,6 +454,261 @@ bool MQ2GroupType::GetMember(MQVarPtr VarPtr, const char* Member, char* Index, M
 			}
 		}
 
+		Dest = pSpawnType->MakeTypeVar();
+		return true;
+	case GroupMembers::Shaman:
+		for(auto& member : *pLocalPC->Group)
```
dannuic, line 460: it's unsafe to deref like this

src/main/datatypes/MQ2GroupType.cpp
```diff
@@ -455,6 +750,64 @@ bool MQ2GroupType::GetMember(MQVarPtr VarPtr, const char* Member, char* Index, M
 			}
 		}
 		return false;
+		
+	case GroupMembers::Furthest:
+		Dest.DWord = 0;
```
dannuic, line 755: No reason to set DWord here, it's redundant

src/main/datatypes/MQ2GroupType.cpp
```diff
@@ -455,6 +750,64 @@ bool MQ2GroupType::GetMember(MQVarPtr VarPtr, const char* Member, char* Index, M
...
+		Dest.Type = pFloatType;
+
+		for(auto& member : *pLocalPC->Group)
+		{
+			if(member && member->pSpawn)
+			{
+				auto distance = Distance3DToSpawn(pLocalPC->pSpawn, member->pSpawn);
```
dannuic, line 762: Since this is 3D, I'd probably call this Furthest3D or something (and the others). Normally we only calculate 2D distance for things like this.

src/main/datatypes/MQ2AdvLootType.cpp
```diff
@@ -345,7 +347,49 @@ bool MQ2AdvLootItemType::GetMember(MQVarPtr VarPtr, const char* Member, char* In
...
+		{
+			strcpy_s(DataTypeTemp, tmp_str.c_str());
```
brainiac, line 359: tmp_str is not necessary throughout this entire diff. There is no reason for it to exist because you can just copy the value you want directly into the destination.

src/main/datatypes/MQ2GroupType.cpp
```diff
@@ -455,6 +750,64 @@ bool MQ2GroupType::GetMember(MQVarPtr VarPtr, const char* Member, char* Index, M
...
+				if(distance > furthest_dist)
+					furthest_dist = distance;
+			}
+		}
+		Dest.Float = furthest_dist;
+		return true;
+
+	case GroupMembers::FurthestMelee:
```
dannuic, line 770: I'm a little confused why these need to be datatypes, this seems really simple to do in scripts

src/main/datatypes/MQ2GroupType.cpp
```diff
@@ -40,6 +40,25 @@ enum class GroupMembers
...
+	Monk,
+	Bard,
+	Rogue,
+	Necromancer,
+	Wizard,
+	Mage,
+	Magician,
+	Enchanter,
+	Beastlord,
+	Berserker,
```
brainiac, line 58: I cannot accept this change as implemented. This is not a pattern I want to propogate -- the same block of code down below is copied 13 times.

`Cleric` is a huge anti pattern that eqmule threw into this because of a personal use case with no bearing on the overall design (case in point).

If it is really important to have this functionality, we need to design a new interface for it that isn't a copy pasta. Perhaps instead we need something like  `MemberByClass`

src/main/datatypes/MQ2GroupType.cpp
```diff
@@ -414,6 +454,261 @@ bool MQ2GroupType::GetMember(MQVarPtr VarPtr, const char* Member, char* Index, M
...
+			if(member
+			   && member->Type == EQP_PC
+			   && member->pSpawn
+			   && member->pSpawn->GetClass() == Berserker)
+			{
+				Dest = pSpawnType->MakeTypeVar(member->pSpawn);
+				return true;
+			}
+		}
+
```
brainiac, line 711: If you ever find yourself copy pasting the same code 13 times, that should be a huge red flag.

src/main/datatypes/MQ2GroupType.cpp
```diff
@@ -455,6 +750,64 @@ bool MQ2GroupType::GetMember(MQVarPtr VarPtr, const char* Member, char* Index, M
...
+				if(class_value <= TotalPlayerClasses && ClassInfo[class_value].PureCaster)
+				{
+					auto distance = Distance3DToSpawn(pLocalPC->pSpawn, member->pSpawn);
+
+					if(distance > furthest_dist)
+						furthest_dist = distance;
+				}
+			}
+		}
+		Dest.Float = furthest_dist;
```
brainiac, line 809: Same issue here. Perhaps we can add parameters to FurthestMember to further limit the result set.

brainiac:
The theme of my response is pretty simple: i don't want to see code copied in more than one place.

Also, it looks like you're using editing files in github. I highly recommend that instead you clone the repository locally and submit your change to your fork in a branch from git

src/main/datatypes/MQ2AdvLootType.cpp
```diff
@@ -232,6 +233,7 @@ MQ2AdvLootItemType::MQ2AdvLootItemType() : MQ2Type("advlootitem")
 	ScopedTypeMember(AdvLootItemMembers, IconID);
 	ScopedTypeMember(AdvLootItemMembers, NoDrop);
 	ScopedTypeMember(AdvLootItemMembers, FreeGrab);
+	ScopedTypeMember( AdvLootItemMembers, Status );
```
brainiac, line 236: whitespace

dannuic (changes requested):
In all honesty, the advloot stuff seems like a useful addition, but I don't know why we need to add more datatype bloat for things that can (and should) be done in scripts (all the group stuff). Is there a compelling reason beyond symmetry to add them?

You should also be aware that your formatting style is inconsistent with the rest of the codebase, so that would have to be fixed before a merge as well.

JBzone:
I think made all the requested changes.

src/main/datatypes/MQ2GroupType.cpp
```diff
@@ -83,26 +65,66 @@ MQ2GroupType::MQ2GroupType() : MQ2Type("group")
...
-	ScopedTypeMember(GroupMembers, Furthest);
-	ScopedTypeMember(GroupMembers, FurthestMelee);
-	ScopedTypeMember(GroupMembers, FurthestPureCaster);
+	ScopedTypeMember(GroupMembers, ClassInGroup);
+	ScopedTypeMember(GroupMembers, FurthestMember);
+}
+
+PlayerClass GetClassIDFromClassName(const char* class_name)
+{
+	static const mq::ci_unordered::map<std::string_view, PlayerClass> class_mapping{
```
brainiac, line 74: `ClassInfo` exists

brainiac:
On the right track for the group member changes, but you actually need to remove the advance loot changes from this PR and put it into a separate one so that the two sets of changes can be reviewed individually.

I think you might be struggling a little bit because you're not actually using git properly with this PR. Theres lots of resources on the internet about how to do this, I picked this one at random which seemed to explain some of the things i would've wanted to explain: https://www.freecodecamp.org/news/how-to-make-your-first-pull-request-on-github-3/

JBzone:
Ya sorry first time using git. I will read up and get this pull request fixed up. Sorry for the jumbled mess!

src/main/datatypes/MQ2GroupType.cpp
brainiac: https://github.com/macroquest/eqlib/blob/live/EQData.h#L51

brainiac:
I integrated the advancedloot changes already, the rest of this still needs some improvement. Its been over a month, is there a plan to address any of the feedback?

JBzone:
Sorry had a death in the family and had to travel so everything outside of
RL got put on the backburner. I will update things soon and reopen a pull
request at a later time.

PR #776, opened 2023-11-10
Add Dev Feature: Persistent commands

Knightly1:
- Persistent logging between sessions (logged to SQLite)
- Should mostly handle multiple clients
- Not intended as an end user feature

### DIFF
--- a/src/main/MQ2ImGuiConsole.cpp
+++ b/src/main/MQ2ImGuiConsole.cpp
@@ -27,6 +27,7 @@
 
 #include "zep.h"
 #include <optional>
+#include "sqlite3.h"
 
 namespace mq {
 
@@ -72,6 +73,7 @@ static bool s_consoleVisible = false;
 static bool s_consoleVisibleOnStartup = false;
 static bool s_resetConsolePosition = false;
 static bool s_setFocus = false;
+static bool s_consoleFileLog = false;
 
 class ImGuiConsole;
 ImGuiConsole* gImGuiConsole = nullptr;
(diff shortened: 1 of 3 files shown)


src/main/MQ2ImGuiConsole.cpp
```diff
@@ -902,6 +904,142 @@ namespace imgui
 	}
 }
 
+std::vector<std::string> initConsoleDatabase(sqlite3*& db, int process_id)
```
brainiac, line 907: InitConsoleDatabase

src/main/MQ2ImGuiConsole.cpp
```diff
@@ -1591,6 +1737,15 @@ static void ConsoleSettings()
 	ImGui::SameLine();
 	mq::imgui::HelpMarker("This feature allows you to automatically show the MacroQuest Console upon load.");
 
+	ImGui::SameLine();
+	if (ImGui::Checkbox("Developer: Command Logging", &s_consoleFileLog))
```
brainiac, line 1741: Command Logging -> Command History

src/main/MQ2ImGuiConsole.cpp
```diff
@@ -1591,6 +1737,15 @@ static void ConsoleSettings()
...
+	{
+		WritePrivateProfileBool("Console", "FileLog", s_consoleFileLog, mq::internal_paths::MQini);
```
brainiac, line 1743: File Log -> Command History

apply that to other places too i suppose

src/main/MQ2ImGuiConsole.cpp
```diff
@@ -902,6 +904,142 @@ namespace imgui
...
+						sqlite3_finalize(stmt);
+					}
+				}
+			}
+		}
+	}
+	return history;
+}
+
+void AddEntryToDatabase(sqlite3*& db, int process_id, const char* entry)
```
brainiac, line 1019: this doesn't appear to write to db, so should'nt be reference?

src/main/MQ2ImGuiConsole.cpp
```diff
@@ -902,6 +904,142 @@ namespace imgui
...
+						WriteChatf("MQ Console Error preparing query for console buffer retrieval: %s", sqlite3_errmsg(db));
+					}
+					else
+					{
+						sqlite3_bind_int(stmt, 1, process_id);
+						sqlite3_bind_int(stmt, 2, process_id);
+
+						while (sqlite3_step(stmt) == SQLITE_ROW)
+						{
+							if (const char* text = reinterpret_cast<const char*>(sqlite3_column_text(stmt, 0)))
```
brainiac, line 1004: what does sqlite3_column_text return? (why is this reinterpret_cast?)

brainiac (changes requested):
Feedback submitted, I like how you resolved different processes. clever solution.

src/main/MQ2ImGuiConsole.cpp
Knightly1: It returns const unsigned char*

src/main/MQ2ImGuiConsole.cpp
Knightly1: History is used internally already -- CommandHistoryLog?

src/main/MQ2ImGuiConsole.cpp
Knightly1: Good catch.  Had it as a reference when I added the menu because I was going to have AddEntryToDatabase init the DB if the checkbox was checked but I didn't like that implementation and backed it out.

PR #786, opened 2023-11-22
Added new WindowType, InvSlotWindow

DerpleMQ2:
Added new WindowType, InvSlotWindow with the following properties:
- Background
- ItemLocation
- ItemIndexSlot1
- ItemIndexSlot2
- ItemIndexSlot3
- Item
- ItemOffsetX
- ItemOffsetY
- ItemTexture
- Quantity
- Selected
- FindSelected
- HotButton
- InventorySlotLinked
- Text
- Mode

Usage:

mq.TLO.Window("BazaarWnd").Child(string.format("BZR_BazaarSlot0")).InvSlot.Selected()

### DIFF
--- a/src/main/datatypes/MQ2WindowType.cpp
+++ b/src/main/datatypes/MQ2WindowType.cpp
@@ -917,9 +917,6 @@ bool MQ2InvSlotWindowType::GetMember(MQVarPtr VarPtr, const char* Member, char* 
 	if (pWnd->GetType() != UI_InvSlot)
 		return false;
 
-	if (!Index || !Index[0])
-		return false;
-
 	CInvSlotWnd* pInvWnd = static_cast<CInvSlotWnd*>(pWnd);
 
 	MQTypeMember* pMember = MQ2InvSlotWindowType::FindMember(Member);
@@ -1017,12 +1014,14 @@ bool MQ2InvSlotWindowType::GetMember(MQVarPtr VarPtr, const char* Member, char* 
 
 bool MQ2InvSlotWindowType::ToString(MQVarPtr VarPtr, char* Destination)
 {
-	CXWnd* pWnd = static_cast<CXWnd*>(VarPtr.Ptr);
+	CInvSlotWnd* pWnd = static_cast<CInvSlotWnd*>(VarPtr.Ptr);
 
 	if (pWnd)
 	{
-		if (pWnd->IsVisible())
-			strcpy_s(Destination, MAX_STRING, "TRUE");
+		ItemPtr pItem = pWnd->LinkedItem ? pWnd->LinkedItem : pLocalPC->GetItemByGlobalIndex(pWnd->ItemLocation);
+
+		if (pItem && pWnd->IsVisible())
+			strcpy_s(Destination, MAX_STRING, pItem->GetName());
 		else
 			strcpy_s(Destination, MAX_STRING, "FALSE");
 	}
@@ -1042,20 +1041,6 @@ bool MQ2InvSlotWindowType::FromString(MQVarPtr& VarPtr, const char* Source)
 {
 	if (VarPtr.Ptr = FindMQ2WindowPath(Source))
 		return true;
-
-	return false;
-}
-
-bool MQ2InvSlotWindowType::dataWindow(const char* szIndex, MQTypeVar& Ret)
-{
-	if (szIndex[0])
-	{
-		if (Ret.Ptr = FindMQ2Window(szIndex))
-		{
-			Ret.Type = pInvSlotWindowType;
-			return true;
-		}
-	}
 
 	return false;
 }
--- a/src/main/datatypes/MQ2WindowType.cpp
+++ b/src/main/datatypes/MQ2WindowType.cpp
@@ -1015,17 +1015,21 @@ bool MQ2InvSlotWindowType::GetMember(MQVarPtr VarPtr, const char* Member, char* 
 bool MQ2InvSlotWindowType::ToString(MQVarPtr VarPtr, char* Destination)
 {
 	CInvSlotWnd* pWnd = static_cast<CInvSlotWnd*>(VarPtr.Ptr);
+	bool bFound = false;
 
 	if (pWnd)
 	{
 		ItemPtr pItem = pWnd->LinkedItem ? pWnd->LinkedItem : pLocalPC->GetItemByGlobalIndex(pWnd->ItemLocation);
 
 		if (pItem && pWnd->IsVisible())
+		{
 			strcpy_s(Destination, MAX_STRING, pItem->GetName());
+			bFound = true;
+		}
 		else
-			strcpy_s(Destination, MAX_STRING, "FALSE");
+			strcpy_s(Destination, MAX_STRING, "NULL");
 	}
-	return true;
+	return bFound;
 }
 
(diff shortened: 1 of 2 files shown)


src/main/datatypes/MQ2WindowType.cpp
```diff
@@ -630,6 +669,15 @@ bool MQ2WindowType::GetMember(MQVarPtr VarPtr, const char* Member, char* Index,
 		return false;
 	}
 
+	case WindowMembers::InvSlot:
+		if (pWnd->GetType() == UI_InvSlot)
+		{
+			Dest.Type = pInvSlotWindowType;
+			Dest.Ptr = VarPtr.Ptr;
```
dannuic, line 676: I'd prefer use of Dest.Set here, this is the old method that we don't need to be using anymore for most things

src/main/datatypes/MQ2WindowType.cpp
```diff
@@ -860,4 +908,156 @@ bool MQ2WindowType::dataWindow(const char* szIndex, MQTypeVar& Ret)
 	return false;
 }
 
+bool MQ2InvSlotWindowType::GetMember(MQVarPtr VarPtr, const char* Member, char* Index, MQTypeVar& Dest)
+{
+	CXWnd* pWnd = static_cast<CXWnd*>(VarPtr.Ptr);
```
dannuic, line 913: You can actually use a getter here that does a lot of checks for you, like this: VarPtr.Get<CXWnd>();

Then you test that for nullptr and UI_InvSlot type

src/main/datatypes/MQ2WindowType.cpp
```diff
@@ -860,4 +908,156 @@ bool MQ2WindowType::dataWindow(const char* szIndex, MQTypeVar& Ret)
...
+	if (!VarPtr.Ptr)
+		return false;
+
+	if (pWnd->GetType() != UI_InvSlot)
+		return false;
+
+	if (!Index || !Index[0])
+		return false;
```
dannuic, line 921: You don't want this here, it will cause all your non-indexed members to fail

src/main/datatypes/MQ2WindowType.cpp
```diff
@@ -860,4 +908,156 @@ bool MQ2WindowType::dataWindow(const char* szIndex, MQTypeVar& Ret)
...
+	{
+		if (Ret.Ptr = FindMQ2Window(szIndex))
+		{
+			Ret.Type = pInvSlotWindowType;
+			return true;
+		}
+	}
+
+	return false;
+}
```
dannuic, line 1061: you don't want this function, it's what's creating the TLO (and this new type doesn't need a TLO)

src/main/datatypes/MQ2WindowType.cpp
```diff
@@ -860,4 +908,156 @@ bool MQ2WindowType::dataWindow(const char* szIndex, MQTypeVar& Ret)
...
+bool MQ2InvSlotWindowType::ToString(MQVarPtr VarPtr, char* Destination)
+{
+	CXWnd* pWnd = static_cast<CXWnd*>(VarPtr.Ptr);
+
+	if (pWnd)
+	{
+		if (pWnd->IsVisible())
+			strcpy_s(Destination, MAX_STRING, "TRUE");
+		else
+			strcpy_s(Destination, MAX_STRING, "FALSE");
```
dannuic, line 1027: This could probably be more useful -- like the item in the slot or something if it exists?

src/main/datatypes/MQ2WindowType.cpp
dannuic: On second thought, this is actually fine -- we don't own the CXWnd pointer, so you'd also need to create a wrapper around it.... not work that needs to be done in this PR

src/main/datatypes/MQ2WindowType.cpp
dannuic: Same as Set, strike this comment because it's more work than you need to do for this PR

src/main/datatypes/MQ2WindowType.cpp
nandor-code: done.

DerpleMQ2: done

src/main/datatypes/MQ2WindowType.cpp
DerpleMQ2: done

dannuic (commented):
it looks good to me, but I think brainiac said he wanted to take a look at it before merging so I'll leave it open for him

brainiac:
I'm going to look at this soon, I promise...

src/main/datatypes/MQ2WindowType.cpp
```diff
@@ -860,4 +908,141 @@ bool MQ2WindowType::dataWindow(const char* szIndex, MQTypeVar& Ret)
...
+	VarPtr.Ptr = Source.Ptr;
+	return true;
+}
+
+bool MQ2InvSlotWindowType::FromString(MQVarPtr& VarPtr, const char* Source)
+{
+	if (VarPtr.Ptr = FindMQ2WindowPath(Source))
+		return true;
+
+	return false;
```
brainiac, line 1045: don't think this function makes sense to have. Its optional, so we can just remove it.

src/main/datatypes/MQ2WindowType.cpp
```diff
@@ -860,4 +908,141 @@ bool MQ2WindowType::dataWindow(const char* szIndex, MQTypeVar& Ret)
...
+	CInvSlotWnd* pWnd = static_cast<CInvSlotWnd*>(VarPtr.Ptr);
+
+	if (pWnd)
+	{
+		ItemPtr pItem = pWnd->LinkedItem ? pWnd->LinkedItem : pLocalPC->GetItemByGlobalIndex(pWnd->ItemLocation);
+
+		if (pItem && pWnd->IsVisible())
+			strcpy_s(Destination, MAX_STRING, pItem->GetName());
+		else
+			strcpy_s(Destination, MAX_STRING, "FALSE");
```
brainiac, line 1026: might be better to go with NULL, or return false.

src/main/datatypes/MQ2WindowType.cpp
```diff
@@ -860,4 +908,141 @@ bool MQ2WindowType::dataWindow(const char* szIndex, MQTypeVar& Ret)
...
+	}
+	return true;
```
brainiac, line 1028: failure case should return false.

src/main/datatypes/MQ2WindowType.cpp
DerpleMQ2: removed.

src/main/datatypes/MQ2WindowType.cpp
DerpleMQ2: changed.

src/main/datatypes/MQ2WindowType.cpp
DerpleMQ2: updated but it does not match the pattern the other windows follow

brainiac: Yeah I thought about this for a while. If you needed/wanted the TRUE/FALSE return for the window (which is non-standard itself), you would just use window type. InvSlotWindow need not adhere to the same pattern IMO

PR #789, opened 2023-11-29
add example for mq luarocks

aquietone:
Add an example script that covers using PackageMan.lua to load each of the available luarocks modules from the MQ luarocks server.


brainiac:
sure

PR #791, opened 2023-12-05
Adding sqlite-backed login profiles

dannuic:


src/login/Login.h
```diff
@@ -41,19 +41,283 @@ struct ProfileRecord
 	std::string hotkey;
 	std::string characterClass;
 	int characterLevel = 0;
+
+	std::optional<std::string> eqPath;
 	bool checked = false;
 
+	std::optional<bool> endAfterSelect;
+	std::optional<int> charSelectDelay;
+	std::optional<std::filesystem::path> customClientIni;
```
brainiac, line 50: this could probably be a string. I think most places i found use it as a string anyways.

src/login/Login.h
```diff
@@ -41,19 +41,283 @@ struct ProfileRecord
...
+		for (auto t : *this)
+			vec.emplace_back(std::move(t));
+		return vec;
+	}
+
+private:
+	std::shared_ptr<WithDb> m_db;
+	StatementHelper m_stmt;
+	std::function<T(sqlite3_stmt*, sqlite3*)> m_result;
+};
```
brainiac, line 223: A lot of this db helper code could probably be put into its own sqlite specific header

src/main/MQPostOffice.cpp
```diff
@@ -77,30 +77,30 @@ class MQPostOffice : public PostOffice
 			case MQMessageId::MSG_ROUTE:
 			{
 				auto envelope = ProtoMessage::Parse<proto::routing::Envelope>(message);
-				const auto& address = envelope.address();
+				const auto& address = envelope.has_address() ? std::make_optional(envelope.address()) : std::nullopt;
```
brainiac, line 80: this shouldn't be const ref anymore.

src/plugins/autologin/MQ2AutoLogin.cpp
```diff
@@ -38,9 +38,11 @@ PreSetup("MQ2AutoLogin");
 
 constexpr int STEP_DELAY = 1000;
 
-fs::path CustomIni;
+std::string DBPath = (fs::path(mq::gPathConfig) / "login.db").string();
```
brainiac, line 41: Do we ever re-evaluate this? can gPathConfig change?

src/plugins/autologin/MQ2AutoLogin.cpp
```diff
@@ -38,9 +38,11 @@ PreSetup("MQ2AutoLogin");
...
 uint64_t ReenableTime = 0;
 postoffice::DropboxAPI s_autologinDropbox;
+bool s_detoursInstalled = false;
```
brainiac, line 44: this is unused

src/plugins/autologin/MQ2AutoLogin.cpp
```diff
@@ -38,9 +38,11 @@ PreSetup("MQ2AutoLogin");
...
+uintptr_t s_joinServer = 0;
```
brainiac, line 45: this isn't static. none of these are static.

src/plugins/autologin/StateMachine.cpp
```diff
@@ -70,78 +69,23 @@ static std::optional<ProfileRecord> UseMQ2Login(CEditWnd* pEditWnd)
...
-		record.profileName = "";
-		record.accountName = account;
-		record.accountPassword = GetPrivateProfileString(account, "Password", "", INIFileName);
-		record.serverName = GetPrivateProfileString(account, "Server", "", INIFileName);
-		record.characterName = GetPrivateProfileString(account, "Character", "", INIFileName);
-		// Override the character select settings if specified
-		Login::m_settings.EndAfterSelect = GetPrivateProfileBool(account, "EndAfterSelect", Login::m_settings.EndAfterSelect, INIFileName);
-		Login::m_settings.CharSelectDelay = GetPrivateProfileInt(account, "CharSelectDelay", Login::m_settings.CharSelectDelay, INIFileName);
+		if (record.customClientIni && record.customClientIni->is_relative())
+			record.customClientIni = std::filesystem::current_path() / *record.customClientIni;
```
brainiac, line 81: don't need to check if path is relative.

src/plugins/autologin/MQ2AutoLogin.cpp
```diff
@@ -429,19 +455,18 @@ void Cmd_Loginchar(SPAWNINFO* pChar, char* szLine)
...
 	}
-	else if (!record.serverName.empty() && !record.accountName.empty() && !record.accountPassword.empty())
+	else if (!record.accountName.empty() && !record.accountPassword.empty())
 	{
-		if (record.characterName.empty())
+		if (record.serverName.empty())
+			LoginServerSelect(
+				record.accountName.c_str(),
+				record.accountPassword.c_str());
+		else if (record.characterName.empty())
```
brainiac, line 469: more of a mental note, should probably convert all this code to use string_view or std::string later.

src/loader/AutoLoginImGui.cpp
```diff
@@ -0,0 +1,2010 @@
...
+
+			ImGui::Spacing();
+			if (ImGui::Checkbox("End After Character Select", &end_after_select.Read()))
+				login::db::WriteSetting("end_after_select", end_after_select.Updated() ? "true" : "false", "Disable the plugin once the character has been selected");
+			ImGui::SetItemTooltip("Disable the plugin once the character has been selected");
+
+			ImGui::Spacing();
+			if (ImGui::Checkbox("Load Legacy Config Next Load", &load_ini.Read()))
+				login::db::WriteSetting("load_ini", load_ini.Updated() ? "true" : "false", "Import data from autologin ini file one time at load");
+			ImGui::SetItemTooltip("Import data from autologin ini file one time at load");
```
brainiac, line 1674: In general these tooltips are hard to discover. They work occasionally but most of the time it is really not clear what has a tooltip and what does not.

Prefer using the HelpMarker function to create a help tooltip. ItemTooltip should really be used in cases where that isn't practical.

src/loader/ImGui.cpp
```diff
@@ -0,0 +1,492 @@
...
+
+			const auto monitor = ImGui::GetViewportPlatformMonitor(ImGui::GetMainViewport());
+			ImGui::SetNextWindowSize({ 640.f, 480.f }, ImGuiCond_FirstUseEver);
+			ImGui::SetNextWindowPos(monitor->WorkPos + (monitor->WorkSize * 0.5f), ImGuiCond_FirstUseEver, { 0.5f, 0.5f });
+
+			if (ImGui::Begin("MacroQuest", &is_open))
+			{
+				static std::optional<const std::function<void()>*> current_callback;
+
+				ImGui::BeginChild("Window Selection", ImVec2(ImGui::GetContentRegionAvail().x * 0.2f, 0), ImGuiChildFlags_Border, ImGuiWindowFlags_HorizontalScrollbar);
```
brainiac, line 376: Check out the ImGuiTreePanelWindow which generalizese this design. The column width should be a fixed size.

src/loader/ImGui.cpp
```diff
@@ -0,0 +1,492 @@
...
+					if (ImGui::Selectable(name.c_str(), is_selected))
+						current_callback = &callback;
+
+					if (is_selected) ImGui::SetItemDefaultFocus();
+				}
+				ImGui::EndChild();
+
+				ImGui::SameLine();
+
+				ImGui::BeginChild("Window Display", ImVec2(0, 0), ImGuiChildFlags_None, ImGuiWindowFlags_HorizontalScrollbar);
```
brainiac, line 389: This ends up being slightly larger than the available space and always creating a horizontal scrollbar for me.

src/loader/AutoLoginImGui.cpp
```diff
@@ -0,0 +1,2010 @@
...
+			
+			ImGui::Spacing();
+			if (ImGui::InputScalar("Character Select Delay", ImGuiDataType_U32, &char_select_delay.Read()))
```
brainiac, line 1677: These text boxes are way too wide. (You could also use InputInt or DragInt).

We should set a fixed width to all the elements in this pane so that it is more readable and doesn't do wierd things when stretched horizontally.

brainiac (changes requested):
Left feedback in discord and throughout the code. Do check this main feed because I think some of the comments went to outdated snippets.

src/login/Login.cpp
```diff
@@ -267,3 +354,1805 @@ void ProfileRecord::FormatTo(char* buffer, size_t length) const
...
+
+	~WithDb()
+	{
+		if (m_db != nullptr) sqlite3_close(m_db);
+	}
+
+	[[nodiscard]] sqlite3* GetDB() const { return m_db; }
+
+	template <typename T>
+	static std::function<T(const std::string&, const std::function<T(sqlite3_stmt*, sqlite3*)>&)> Query(int flags)
```
brainiac, line 388: Why are we currying flags? Why not just take the query and action as parameters?

src/login/Login.cpp
```diff
@@ -267,3 +354,1805 @@ void ProfileRecord::FormatTo(char* buffer, size_t length) const
...
+	}
+
+	WithDb(const WithDb&) = delete;
+	WithDb(WithDb&&) = delete;
+	WithDb& operator=(const WithDb&) = delete;
+	WithDb& operator=(WithDb&&) = delete;
+
+	static const std::shared_ptr<WithDb>& Get(int flags)
+	{
+		static std::map<int, std::shared_ptr<WithDb>> connections;
```
brainiac, line 404: how do you remove connections?

src/loader/AutoLoginImGui.cpp
```diff
@@ -0,0 +1,2010 @@
...
+			ImGui::TableSetupColumn("Server");
+			ImGui::TableSetupColumn("Character");
+			ImGui::TableSetupColumn("Persona");
+			ImGui::TableSetupColumn("Hotkey");
+			ImGui::TableSetupColumn("##play");
+			ImGui::TableHeadersRow();
+
+			for (auto& profile : login::db::GetProfiles(info.profileName))
+			{
+				ImGui::PushID(&profile);
```
brainiac, line 1250: This PushID isn't working because profile is a value type and is always occupying the same space on the stack. The reference and auto is misleading.

Best solve I have is to just count rows with an index.

brainiac: The reference actually makes this const reference. Which makes profile immutable, which I eblieve you intended for Selectable down below to mutate the checked flag. (otherwise the selectable doesn't make sense -- how do you toggle the checked state?)

src/loader/AutoLogin.cpp
```diff
@@ -140,1906 +108,1987 @@ static SClassInfo ClassInfo[] =
...
-const char* GetClassShortName(DWORD playerclass)
+const char* GetClassShortName(const DWORD player_class)
 {
-	return ClassInfo[playerclass].UCShortName;
+	return s_classInfo[player_class].UCShortName;
 }
 
-void SelectCharacterClass(const std::string& cclass)
+namespace {
+// This sends a message to self so that we ensure that all access of s_loadedInstances happens on the main thread
```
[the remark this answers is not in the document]
dannuic: This replaces a function that was being called from windows messaging threads, but all of that has been removed and replaced by imgui. I thought it better to keep the messaging in

src/loader/imgui_backend/imgui_engine.cpp
```diff
@@ -0,0 +1,207 @@
...
+	sd.Windowed = true;
+	sd.SwapEffect = DXGI_SWAP_EFFECT_DISCARD;
+
+	uint32_t createDeviceFlags = 0;
+	D3D_FEATURE_LEVEL featureLevel;
+	const D3D_FEATURE_LEVEL featureLevelArray[2] = { D3D_FEATURE_LEVEL_11_0, D3D_FEATURE_LEVEL_10_0, };
+	HRESULT res = D3D11CreateDeviceAndSwapChain(nullptr, D3D_DRIVER_TYPE_HARDWARE, nullptr, createDeviceFlags, featureLevelArray, 2, D3D11_SDK_VERSION, &sd, &s_swapChain, &s_d3dDevice, &featureLevel, &s_d3dDeviceContext);
+	if (res == DXGI_ERROR_UNSUPPORTED)
+		res = D3D11CreateDeviceAndSwapChain(nullptr, D3D_DRIVER_TYPE_WARP, nullptr, createDeviceFlags, featureLevelArray, 2, D3D11_SDK_VERSION, &sd, &s_swapChain, &s_d3dDevice, &featureLevel, &s_d3dDeviceContext);
+	if (res != S_OK)
```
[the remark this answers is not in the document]
dannuic: I've tested with a KVM, which should cover that

src/loader/ImGui.cpp
```diff
@@ -0,0 +1,492 @@
...
+#include "imgui_backend/imgui_engine.h"
+
+#include "imgui/imgui.h"
+#include "imgui/imgui_internal.h"
+
+#include "spdlog/spdlog.h"
+
+namespace {
+// map of panels in the main GUI window
+std::map<std::string, std::function<void()>> s_panels;
```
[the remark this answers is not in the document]
dannuic: I added a way to select a panel from outside -- the string is needed now

src/loader/ImGui.cpp
```diff
@@ -0,0 +1,492 @@
...
+
+// map of viewport windows to render
+std::map<std::string, std::function<bool()>> s_viewports;
```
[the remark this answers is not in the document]
dannuic: This prevents the same viewport from getting opened twice

src/loader/ImGui.cpp
```diff
@@ -0,0 +1,492 @@
...
+{
+	float Alpha = 0.f;
+	ImGuiViewportFlags OriginalFlagsSet = ImGuiViewportFlags_None;
+	ImGuiViewportFlags OriginalFlagsClear = ImGuiViewportFlags_None;
+	ImGuiID ParentViewportID = 0;
+};
+
+std::map<std::string, Modal> s_modals;
+
+struct DimRatio
```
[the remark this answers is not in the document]
dannuic: It's a storage class for dim ratios on viewports (note s_dimRatios is a map of imgui ID to DimRatio)

src/loader/ImGui.cpp
```diff
@@ -0,0 +1,492 @@
...
+{
+	float Ratio = 0.f;
+	bool Open = true;
+};
+
+std::map<ImGuiID, DimRatio> s_dimRatios;
```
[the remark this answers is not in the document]
dannuic: this can't be a vector, I need the ID to know which viewports to dim in the background

src/loader/ImGui.cpp
```diff
@@ -0,0 +1,492 @@
...
+
+struct Modal
+{
+	float Alpha = 0.f;
+	ImGuiViewportFlags OriginalFlagsSet = ImGuiViewportFlags_None;
+	ImGuiViewportFlags OriginalFlagsClear = ImGuiViewportFlags_None;
+	ImGuiID ParentViewportID = 0;
+};
+
+std::map<std::string, Modal> s_modals;
```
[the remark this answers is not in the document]
dannuic: again, need a map here

src/loader/ImGui.h
```diff
@@ -0,0 +1,54 @@
...
+#pragma once
+
+#include <functional>
+#include <string>
+
+#include "imgui/imgui.h"
+
+#include <windows.h>
+
+typedef void (_cdecl * SDL_WindowsMessageHook)(void *userdata, void *hWnd, unsigned int message, uint64_t wParam, int64_t lParam);
```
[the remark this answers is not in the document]
dannuic: it is indeed, I forgot to remove then when I removed the SDL backend

src/plugins/autologin/MQ2AutoLogin.cpp
dannuic: hm, my assumption has always been that this cannot change because all plugins treat it as static as part of the plugin API. I think if this mutates, there will be lots of issues outside of this

src/login/Login.cpp
dannuic: at one point, I was storing this function per flags, but when the interface changed I never changed it

src/loader/ImGui.cpp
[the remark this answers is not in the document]
brainiac: can still be a vector

src/login/Login.cpp
dannuic: they get shut down gracefully when the object dies, but I'll add a Close static to do that

src/loader/ImGui.cpp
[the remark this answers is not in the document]
brainiac: can stil be a vector (or pairs). besides, how would the same viewport get added twice?

src/loader/ImGui.cpp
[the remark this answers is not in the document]
brainiac: so a not-very-bright ratio?

src/loader/ImGui.cpp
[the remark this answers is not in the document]
brainiac: can absolutely be a vector. did you know that std::map is actually a set? If it can be a set it can be a vector too. (of pairs)

src/loader/ImGui.cpp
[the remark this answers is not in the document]
brainiac: need is a strong word.

src/login/Login.cpp
brainiac: its a shared_ptr stored in a function-level static. it will never die.

src/loader/AutoLoginImGui.cpp
```diff
@@ -0,0 +1,1931 @@
...
+						selected.Character.Server = profile.serverName;
+						remove_message = fmt::format("Are you sure you want to remove profile '{}'?", info.Preview());
+						LauncherImGui::OpenModal("Remove Profile");
+					}
+
+					ImGui::EndPopup();
+				}
+
+				ImGui::SameLine();
+				ImGui::Text(profile.accountName.c_str());
```
brainiac, line 1257: These Calls to ImGui::Text need %s or TextUnformatted

src/plugins/autologin/MQ2AutoLogin.cpp
```diff
@@ -38,9 +38,10 @@ PreSetup("MQ2AutoLogin");
 
 constexpr int STEP_DELAY = 1000;
 
-fs::path CustomIni;
+std::string DBPath = (fs::path(mq::gPathConfig) / "login.db").string();
 uint64_t ReenableTime = 0;
 postoffice::DropboxAPI s_autologinDropbox;
+uintptr_t s_joinServer = 0;
```
brainiac, line 44: needs `static`. The s_ means they are static, but they aren't...

src/loader/ImGui.cpp
```diff
@@ -0,0 +1,491 @@
...
+	{
+		ImGuiWindowClass cl;
+		cl.ViewportFlagsOverrideSet |= ImGuiViewportFlags_TopMost;
+		cl.ViewportFlagsOverrideClear |= ImGuiViewportFlags_NoFocusOnAppearing | ImGuiViewportFlags_NoFocusOnClick;
+		cl.ParentViewportId = 0;
+		cl.ClassId = 1;
+		return cl;
+	}();
+
+	struct Modal
```
brainiac, line 49: everything in this file got indented at this point forward.

src/loader/AutoLoginImGui.cpp
```diff
@@ -0,0 +1,1931 @@
...
+					remove_message = fmt::format("Are you certain you want to remove character '{} : {}'? All associated profiles will also be removed.", match.characterName, match.serverName);
+					selected_profile = match;
+					LauncherImGui::OpenModal("Remove Character");
+				}
+
+				ImGui::EndPopup();
+			}
+
+			ImGui::SameLine();
+			ImGui::Text(match.accountName.c_str());
```
brainiac, line 967: same here, need %s or TextUnformatted

brainiac (approved):

PR #794, opened 2023-12-10
Finish DepositItem implementation for Tradeskilldepot window methods

jessebevil:
Complete the implementation for DepositItem method for TradeskillDepotType

Since using /notify methods on TD_Item_List child results in simply clicking on the center item in the listbox, and simply issuing leftmousedown/up doesn't deposit the item, and the button that captures the deposit of an item isn't available in the xml document to notify, the consensus has been to get the center point, mouseto, then just left click in place. So I've implemented this in the method.

### DIFF
--- a/src/main/datatypes/MQ2TradeskillDepotType.cpp
+++ b/src/main/datatypes/MQ2TradeskillDepotType.cpp
@@ -156,8 +156,7 @@ bool MQ2TradeskillDepotType::GetMember(MQVarPtr VarPtr, const char* Member, char
 				if (pCursorAttachment && pCursorAttachment->Type == eCursorAttachment_Item)
 				{
 					CXPoint pt = pTradeskillDepotWnd->pItemList->GetClientRect().CenterPoint();
-					MoveMouse(pt.x, pt.y);
-					Click(pLocalPlayer, "left");
+					MoveMouse(pt.x, pt.y, true);
 				}
 			}
 #endif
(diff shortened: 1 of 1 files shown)


src/main/datatypes/MQ2TradeskillDepotType.cpp
```diff
@@ -148,22 +148,22 @@ bool MQ2TradeskillDepotType::GetMember(MQVarPtr VarPtr, const char* Member, char
...
+		case TradeskillDepotTypeMethods::DepositItem:
+		{
+#if HAS_TRADESKILL_DEPOT
+			if (pTradeskillDepotWnd->IsVisible() && pActiveBanker)
+			{
+				if (pCursorAttachment && pCursorAttachment->Type == eCursorAttachment_Item)
+				{
+					CXPoint pt = pTradeskillDepotWnd->pItemList->GetClientRect().CenterPoint();
+					sprintf_s(DataTypeTemp, "%d %d", pt.x, pt.y);
+					MouseTo(pLocalPlayer, DataTypeTemp);
```
brainiac, line 160: Use `MoveMouse` instead of formatting a string and calling `MouseTo`

jessebevil: Updated

brainiac:
Thanks, I took this and manually applied it in 543725009b74e582d6c4d7fcce956c959210d55a

PR #795, opened 2023-12-13
Add LS currencies. LaurionInnVoucher and ShalowainsPrivateReserve

jessebevil:


brainiac:
`Me.AltCurrency` so that we don't have to keep adding new data members to character...

PR #799, opened 2023-12-18
Add old plugin events to lua scripts by convention

projecteon:
Have a need for some of these plugin events, and this hooks straight into current lua scripts
Had a few different options for implementation but ended on the simpler one which just tries to hook into convention based named global functions in the startup script

Supports following global functions in startup script OnAddSpawn = function (spawn)
OnRemoveSpawn = function (spawn)
OnAddGroundItem= function (groundItem)
OnRemoveGroundItem= function (groundItem)
OnBeginZone= function ()
OnEndZone= function ()
OnZoned= function ()

ie
```lua
local mq = require 'mq'

OnAddSpawn = function (spawn)
    print("Add spawn: "..spawn.ID().." "..spawn.Name())
end

OnRemoveSpawn = function (spawn)
    print("Remove spawn: "..spawn.ID().." "..spawn.Name())
end

OnBeginZone= function (OnBeginZone)
    print("OnBeginZone")
end

while true do
    mq.delay(50)
end
```

### DIFF
--- a/src/plugins/lua/MQ2Lua.cpp
+++ b/src/plugins/lua/MQ2Lua.cpp
@@ -1763,11 +1763,15 @@ PLUGIN_API void InitializePlugin()
 	s_pluginInterface = new LuaPluginInterfaceImpl();
 
 	bindings::InitializeBindings_MQMacroData();
+	
+	LuaActors::Start();
 }
 
 PLUGIN_API void ShutdownPlugin()
 {
 	using namespace mq::lua;
+	
+	LuaActors::Stop();
 
 	bindings::ShutdownBindings_MQMacroData();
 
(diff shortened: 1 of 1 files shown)


brainiac:
This puts a lot of load on the lua engine to be constantly hammering every script for events they probably do not have. We have a branch thats got some work in progress flavor of what you're trying to do here, i'm hoping to pick it back up in january, maybe.

https://github.com/macroquest/macroquest/tree/lua_plugin

projecteon:
This code isnt meant to be acceptance ready but to give an idea on how to solve it, then clean it up for "production" if the pattern is accepted

projecteon:
> This puts a lot of load on the lua engine to be constantly hammering every script for events they probably do not have. We have a branch thats got some work in progress flavor of what you're trying to do here, i'm hoping to pick it back up in january, maybe.
> 
> https://github.com/macroquest/macroquest/tree/lua_plugin

It only calls to the scripts that has a valid hook on startup.
This solution has its own drawbacks, and I discussed with Dannuic on doing another approach (just havent had the time to look into it)

Ill dig into that branch, but my take was to now have a separate "lua plugin script" but let the current scripts hook into the "pulse events"

PR #803, opened 2024-01-05
Updated TargetInfoPH for LS

FelisMalum:


brainiac (changes requested):
You're going to need to do this without rewriting the whole file

FelisMalum:
Still figuring this out, aparently uploading the file wasnt the best idea, will cancel request and just append the original.

brainiac:
github file upload isn't really good because it does this. you'll have better luck using git

PR #804, opened 2024-01-05
Updated for LS (FelisMalum)

FelisMalum:
That looks allot more like I originally wanted.


FelisMalum:
What I did, it was a mess so I just deleted the fork and recreated it with just the added changes so only a +67ish instead of like 1300

PR #805, opened 2024-01-06
ImGuiConsole now works with clipping... don't ask.

DerpleMQ2:
ImGuiConsole has a new property called "opacity" that you can set on creation which will multiply all theme attributes with that alpha value.

### DIFF
--- a/src/main/ImGuiZepEditor.cpp
+++ b/src/main/ImGuiZepEditor.cpp
@@ -127,8 +127,8 @@ private:
 	virtual void DrawRectFilled(const NRectf& rc, ZepColor col) const override;
 	virtual void SetClipRect(const NRectf& rc) override { m_clipRect = rc; }
 	virtual Zep::ZepFont& GetFont(ZepTextType type) override;
-	
-	private:
+
+private:
 	float m_lastClick = -1.0f;
 	bool m_mouseDrag = false;
 	bool m_mouseInside = false;
--- a/src/main/ImGuiZepEditor.cpp
+++ b/src/main/ImGuiZepEditor.cpp
@@ -127,8 +127,8 @@ private:
 	virtual void DrawRectFilled(const NRectf& rc, ZepColor col) const override;
 	virtual void SetClipRect(const NRectf& rc) override { m_clipRect = rc; }
 	virtual Zep::ZepFont& GetFont(ZepTextType type) override;
-
-private:
+	
+	private:
 	float m_lastClick = -1.0f;
 	bool m_mouseDrag = false;
 	bool m_mouseInside = false;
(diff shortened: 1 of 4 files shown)


src/main/ImGuiZepEditor.cpp
```diff
@@ -127,8 +127,8 @@ class ZepEditor_ImGui : public Zep::ZepDisplay, public Zep::ZepEditor
 	virtual void DrawRectFilled(const NRectf& rc, ZepColor col) const override;
 	virtual void SetClipRect(const NRectf& rc) override { m_clipRect = rc; }
 	virtual Zep::ZepFont& GetFont(ZepTextType type) override;
-
-private:
+	
+	private:
```
brainiac, line 131: revert this diff

src/main/ImGuiZepEditor.cpp
```diff
@@ -171,7 +171,8 @@ void ZepEditor_ImGui::DrawChars(ZepFont& font, const NVec2f& pos, ZepColor col,
 	}
 	else
 	{
-		drawList->PushClipRect(toImVec2Adjusted(m_clipRect.topLeftPx, m_screenPos), toImVec2Adjusted(m_clipRect.bottomRightPx, m_screenPos));
+		//drawList->PushClipRect(toImVec2Adjusted(m_clipRect.topLeftPx, m_screenPos), toImVec2Adjusted(m_clipRect.bottomRightPx, m_screenPos));
+		drawList->PushClipRect(toImVec2Adjusted(m_clipRect.topLeftPx, m_screenPos), toImVec2Adjusted(m_clipRect.bottomRightPx, m_screenPos), true);
```
brainiac, line 175: remove commented out code

src/main/MQ2ImGuiConsole.cpp
```diff
@@ -592,6 +605,7 @@ struct ImGuiZepConsole : public mq::imgui::ConsoleWidget, public mq::imgui::ImGu
 	std::shared_ptr<ZepConsoleSyntax> m_syntax;
 	int m_maxBufferLines = 10000;
 	bool m_autoScroll = true;
+	float m_alphaOverride = 1.0f;
```
brainiac, line 608: unused

src/main/MQ2ImGuiConsole.cpp
```diff
@@ -892,6 +906,13 @@ struct ImGuiZepConsole : public mq::imgui::ConsoleWidget, public mq::imgui::ImGu
 	{
 		m_maxBufferLines = maxBufferLines;
 	}
+
+	float GetOpacity() const override { return m_theme->GetOpacity(); }
+
+	void SetOpacity(float opacity) override
+	{
+		m_theme->SetOpacity(opacity);
+	}
```
brainiac, line 915: these are overrides but i don't see the base class that its overriding

brainiac (changes requested):

src/main/ImGuiZepEditor.cpp
DerpleMQ2: fixed.

src/main/MQ2ImGuiConsole.cpp
DerpleMQ2: fixed

PR #809, opened 2024-01-15
Add lua eval console menu item

aquietone:
Any thoughts on something like this? Want to put it next to macro expression eval but don't want to go rewrite the thing in C++.


brainiac:
what if... I converted the lua script into c++?

brainiac:
Haven't really thought about it much, but I think it would make the most sense to be implemented in the lua plugin. That would require some way to hook into the menu, yeah.

I need to think about that whole menu system because i haven't really put a lot of effort into it.

aquietone:
<https://github.com/aquietone/misclua/blob/main/eval.lua>
<https://www.redguides.com/community/resources/lua-expression-evaluator.2589/>

PR #810, opened 2024-01-16
Fix missing apostrophe in LS Raid Currency

Sicprofundus:


Knightly1:
I'm going to close this.  It wouldn't be backwards compatible with the previous setting (typo) which has been around for a month.  It is superseded by #814

PR #814, opened 2024-01-19
Use dbstr for expansions and alt currencies

Knightly1:
- Add remove_chars for removing specific characters in a string
- Deprecate szZoneExpansionName (and update for Laurion's Song)
- Update GetZoneExpansionX to use dbstr for expansions
- Update Alternate Currency Enums to match dbstr consistently
- Modify GetCurrencyIDByName to use dbstr for alt currencies -- this will allow for mq.TLO.Me.AltCurrency('<name>')() to work with the singular or plural name of the currency.  Additionally, it will match regardless of apostrophe or backtick to be compatible with previous changes

### DIFF
--- a/src/main/MQ2Utilities.cpp
+++ b/src/main/MQ2Utilities.cpp
@@ -1310,58 +1310,31 @@ int GetLanguageIDByName(const char* szName)
 	return -1;
 }
 
-// TOOD: Convert to data table
-int GetCurrencyIDByName(char* szName)
+int GetCurrencyIDByName(const char* szName)
 {
-	if (!_stricmp(szName, "Doubloons")) return ALTCURRENCY_DOUBLOONS;
-	if (!_stricmp(szName, "Orux")) return ALTCURRENCY_ORUX;
-	if (!_stricmp(szName, "Phosphenes")) return ALTCURRENCY_PHOSPHENES;
-	if (!_stricmp(szName, "Phosphites")) return ALTCURRENCY_PHOSPHITES;
-	if (!_stricmp(szName, "Faycitum")) return ALTCURRENCY_FAYCITES;
-	if (!_stricmp(szName, "Chronobines")) return ALTCURRENCY_CHRONOBINES;
-	if (!_stricmp(szName, "Silver Tokens")) return ALTCURRENCY_SILVERTOKENS;
-	if (!_stricmp(szName, "Gold Tokens")) return ALTCURRENCY_GOLDTOKENS;
-	if (!_stricmp(szName, "McKenzie's Special Brew")) return ALTCURRENCY_MCKENZIE;
-	if (!_stricmp(szName, "Bayle Marks")) return ALTCURRENCY_BAYLE;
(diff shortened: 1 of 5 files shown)


Knightly1:
Just a note on this for whoever is reviewing.  Decide whether `GetZoneExpansionName(0)` needs to be backwards compatible and return "Original EQ".  It currently would return "EverQuest"

src/main/MQ2Utilities.cpp
```diff
@@ -1310,58 +1310,31 @@ int GetLanguageIDByName(const char* szName)
...
+			{
+				if (ci_equals(currency, remove_chars(currency_name_singular, chars_to_remove)))
+					return i;
+			}
+		}
+	}
+
+	// Crowns sit outside the ALTCURRENCY_LAST range
+	if (ci_equals(currency, "Crowns") || ci_equals(currency, "Crown"))
+		return ALTCURRENCY_CROWNS;
```
brainiac, line 1336: Given the amount of work this function is doing, I would expect that we build a lookup table first.

Knightly1: Implemented a cache as a gut check.  Concerns would be that there's no eviction right now.  I could go down the road of implementing a cache class that tracks LFU...buuuut...

brainiac: This doesn't really solve my concern. All the known values to compare about are known and do not change, so I'd prefer we build a lookup table and then only ever use that.

With this implementation, we will always call remove_chars on the input, and remove_chars on every string that we didn't already match, when that work is redundant.

just build the lookup table, and check if parameter is in the lookup table.

Knightly1: Something like this? 

```cpp
void UpdateCurrencyCache(std::unordered_map<std::string, int>& cache, int value, eDatabaseStringType type)
{
	constexpr std::string_view chars_to_remove = "'`";
	if (const char* ptr = pCDBStr->GetString(value, type))
	{
		const std::string currency = to_lower_copy(ptr);
		cache[currency] = value;
		cache[remove_chars(currency, chars_to_remove)] = value;
	}
}

int GetCurrencyIDByName(const char* szName)
{
	static std::unordered_map<std::string, int> cache;

	if (cache.empty())
	{
		for (int i = ALTCURRENCY_FIRST; i <= ALTCURRENCY_LAST; ++i)
		{
			UpdateCurrencyCache(cache, i, eAltCurrencyNamePlural);
			UpdateCurrencyCache(cache, i, eAltCurrencyName);
		}
		// Crowns are outside ALTCURRENCY_LAST
		UpdateCurrencyCache(cache, ALTCURRENCY_CROWNS, eAltCurrencyNamePlural);
		UpdateCurrencyCache(cache, ALTCURRENCY_CROWNS, eAltCurrencyName);
	}

	const auto it = cache.find(to_lower_copy(szName));
	if (it != cache.end())
	{
		return it->second;
	}

	return -1;
}
```

brainiac: yep

Knightly1: Updated

PR #823, opened 2024-01-29
Added mq.TLO.Me.AbilityReuse which returns the total time an abilit…

DerpleMQ2:
Added mq.TLO.Me.AbilityReuse which returns the total time an ability takes to refresh after it has been used

PR #824, opened 2024-01-31
Ability work

### LINKED ISSUE #371: Add HasInnateSkill for Skill checking
In GitLab by @sicprofundus on Aug 13, 2021, 02:54

in constant.h we have

```
constexpr int NUM_SKILLS = 100;
```
from in-game if we do a /skills
we get an output of all the skills and our skill level

you can see here that slam shows up at 111

image

this makes it so if we do a `/doability slam` mq2 says:

"you do not have this skill (slam)"

image

where other abilities - like begging, work as expected.

it doesn't appear to be simply a matter of adjusting NUM_SKILLS because we get this if we change the value
"Size of PcProfile does not match PcProfile_size"

MQ2Melee appears to have it checked specifically here with the "111"
```cpp
int SKReady(unsigned long id) {
    if (id < 100 || id == 111 || id == 114 || id == 115 || id == 116)
    {
        return pSkillMgr->IsAvailable(id);
    }
    if (id == 105 || id == 107) return LoH_HT_Ready();
    return false;
}
```

not really sure where to go from here otherwise i'd try and have done an MR

edit:

to add further information - it isn't something that is actually listed as a skill with a value

image

yet it says it is a skill

image

/doability in mq2commands has, what appears to be, some handling for abilities like this if they are innate skills, but doesn't appear to function as indicated by the pics above

our skills.h file has all the way up to 132 (Throw Stone) and has the slam at 111

`"Slam",                    // 111`

Knightly1:
- Move IsNumber to strings.h
- Update skills.h with innate skill names
- Update szInnates with appropriate data
- Split Ability functions into their own files and add documentation User Facing:
- /doability will now accept quoted or unquoted ability names
- /doability will now work for innate skills (Fixes #371)
- mq.TLO.Me.Ability is now a boolean. It returns true or false based on whether you have the ability.
- mq.TLO.Me.AbilityReady will no longer say an ability is ready if you don't have that ability.
- Added mq.TLO.Me.AbilityTimerTotal which will return the total amount of time an ability takes to refresh. This is only available while the ability is in cooldown, otherwise it returns 0. Useful for converting mq.TLO.Me.AbilityTimer into a percentage.  (Supersedes #823)

### DIFF
--- a/include/mq/api/Abilities.h
+++ b/include/mq/api/Abilities.h
@@ -1,0 +1,200 @@
+/*
+ * MacroQuest: The extension platform for EverQuest
+ * Copyright (C) 2002-2023 MacroQuest Authors
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License, version 2, as published by
+ * the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ */
+
+#pragma once
+
+#include "eqlib/Globals.h"
+
+using namespace eqlib;
+
+namespace mq {
+
+/**
+ * @fn GetAdjustedSkill
+ *
+ * @brief Retrieves the adjusted skill level for a given skill
+ *
+ * Returns the total skill level for a given skill, including
+ * any bonuses from items, spells, or other sources.
+ *
+ * @param nSkill The skill identifier
+ *
+ * @return int The adjusted skill level or 0 if pLocalPC is not available.
+ **/
+MQLIB_API int GetAdjustedSkill(int nSkill);
+
+/**
+ * @fn GetBaseSkill
+ *
+ * @brief Retrieves the base skill level for a given skill
+ *
+ * Returns the base skill level for a given skill, before
+ * any bonuses from items, spells, or adjustments.
+ *
+ * @param nSkill The skill identifier
+ *
+ * @return int The base skill level or 0 if pLocalPC is not available.
+ **/
+MQLIB_API int GetBaseSkill(int nSkill);
+
+/**
+ * @fn HasSkill
+ *
+ * @brief Checks if the character has a specific skill
+ *
+ * Returns whether the character has a specific skill. Does not
+ * work for innate skills.
+ *
+ * @param nSkill The skill identifier
+ *
+ * @return bool True if the character has the specified skill, false otherwise
+ **/
+MQLIB_API bool HasSkill(int nSkill);
+
+/**
+ * @fn SkillIsActivatable
+ *
+ * @brief Checks if the skill is activatable
+ *
+ * Returns whether the skill is activatable. An activatable skill
+ * is also known as an Ability. Does not include innate skills.
+ *
+ * @param nSkill The skill identifier
+ *
+ * @return bool True if the skill is activatable, false otherwise
+ **/
+MQLIB_API bool SkillIsActivatable(int nSkill);
+
+/**
+ * @fn HasInnateSkill
+ *
+ * @brief Checks if the character has a specific innate skill
+ *
+ * Returns whether the character has a specific innate skill. Does not
+ * work for non-innate skills. Innate skills can be specified as either
+ * their actual number or the "stacked" number.
+ *
+ * For example, slam is identified by both 111 and 11.
+ *
+ * @param nSkill The innate skill identifier (either version)
+ *
+ * @return bool True if the character has the specified innate skill, false otherwise
+ **/
+MQLIB_API bool HasInnateSkill(int nSkill);
+
+/**
+ * @fn InnateSkillIsActivatable
+ *
+ * @brief Checks if the innate skill is activatable
+ *
+ * Returns whether the innate skill is activatable. An activatable skill
+ * is also known as an Ability. Does not work for non-innate skills.
+ * Innate skills can be specified as either their actual number or the
+ * "stacked" number.
+ *
+ * @param nSkill The innate skill identifier (either version)
+ *
+ * @return bool True if the innate skill is activatable, false otherwise
+ **/
+MQLIB_API bool InnateSkillIsActivatable(int nSkill);
+
+/**
+ * @fn HasSkillOrInnate
+ *
+ * @brief Checks if the character has a specific skill
+ *
+ * Returns whether the character has a specific skill. This includes
+ * innate skills, but innate skills must be specified as their "stacked"
+ * (above 100) number.
+ *
+ * @param nSkill The skill identifier
+ *
+ * @return bool True if the character has the specified skill, false otherwise
+ **/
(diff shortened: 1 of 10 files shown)


include/mq/api/Abilities.h
```diff
@@ -0,0 +1,200 @@
...
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ */
+
+#pragma once
+
+#include "eqlib/Globals.h"
```
brainiac, line 17: i don't think we actually need anything from eqlib in this header. instead we need string_view

brainiac: and mq/base/Common.h

Knightly1: Ah, yeah, I had that because I started this as header only.  I can move the using eqlib into the cpp.

src/main/MQ2Commands.cpp
```diff
@@ -2670,41 +2656,28 @@ void DoAbility(SPAWNINFO* pChar, char* szLine)
...
 				{
-					WriteChatf("you do not have this skill (%s)", szBuffer);
-					return;
+					if (SkillOrInnateIsActivatable(Index))
+					{
+						pLocalPC->UseSkill(static_cast<uint8_t>(Index), pLocalPlayer);
+					}
+					else
+					{
+						WriteChatf("DoAbility: Skill is not activatable (%s)", szBuffer);
```
brainiac, line 2674: This message is not consistent with the others ("DoAbility:")

src/main/api/Abilities.cpp
```diff
@@ -0,0 +1,141 @@
...
+
+#include "pch.h"
+
+#include "mq/api/Abilities.h"
+
+#include "eqlib/PcClient.h"
+#include "eqlib/EQClasses.h"
+
+// Defined in MQ2Globals.cpp but we don't want to bring in all of MQ2Globals.h
+MQLIB_VAR const char* szInnates[];
```
brainiac, line 23: This isn't defined in the correct namespace.

src/main/MQ2Commands.cpp
Knightly1: If I make the others consistent, I potentially break any events.  I can make this one consistent with the others but there's no indication where the message came from.

src/main/api/Abilities.cpp
```diff
@@ -0,0 +1,141 @@
...
+
+bool HasSkill(int nSkill)
+{
+	return pLocalPC && pLocalPC->HasSkill(nSkill);
+}
+
+bool SkillIsActivatable(int nSkill)
+{
+	return nSkill < NUM_SKILLS && pSkillMgr && pSkillMgr->pSkill[nSkill]->Activated;
+}
```
brainiac, line 45: these could be inline?

src/main/api/Abilities.cpp
```diff
@@ -0,0 +1,141 @@
...
+	// Allow for both
+	if (nSkill >= NUM_SKILLS)
+		nSkill -= NUM_SKILLS;
+
+	if (nSkill < 0 || nSkill > NUM_INNATE)
+		return false;
+
+	const PcProfile* pProfile = pLocalPC ? pLocalPC->GetCurrentPcProfile() : nullptr;
+	// 255 is the default value for an unset innate skill
+	return 	pProfile && pProfile->InnateSkill[nSkill] != 255;
```
brainiac, line 59: PcProfile is always going to exist if pLocalPC exists, so this could be simplified.

include/mq/api/Abilities.h
```diff
@@ -0,0 +1,200 @@
...
+ * this will return a failure.
+ *
+ * @see GetSkillIDFromString
+ *
+ * @param svString The string representing the ability, either as a name or a number
+ * @param iReturnOnFail (Optional) The value to return on failure to convert, defaults to -1
+ *
+ * @return int The ability identifier corresponding to the string, or iReturnOnFail on failure
+ **/
+MQLIB_API int GetAbilityIDFromString(std::string_view svString, int iReturnOnFail = -1);
```
brainiac, line 198: GetAbilityIDFromString or GetAbilityIDByString? hmm..

src/main/MQ2Commands.cpp
brainiac: That seems like its already a problem for the rest of this function, then? Its no worse off than it already is.

brainiac:
My biggest concerns come from the composition of the header. inline vs exports. Inline has the niceities of not requiring more dll exports. exports mean leaner headers.

but we're already including the whole universe in anything that would see this plugin so i'm tending to lean towards the simpler things being inline.

thoughts?

Knightly1:
> My biggest concerns come from the composition of the header. inline vs exports. Inline has the niceities of not requiring more dll exports. exports mean leaner headers.
> 
> but we're already including the whole universe in anything that would see this plugin so i'm tending to lean towards the simpler things being inline.
> 
> thoughts?

Biggest reason that I leaned towards exports on this one is I'm not sure we're done here and I don't know how complex we're going to get with Abilities.  If this was everything, inline, no question.  But I don't think it's everything.  Of course, we can always inline now and split out later.

brainiac:
we can inline the smaller functions and go half way then.

src/main/MQ2Commands.cpp
Knightly1: Six of one, but I prefer the inconsistency and clarity on the updated function over the consistency without clarity.

brainiac: Well then it should say /doability then, right? DoAbility isn't meaningful to anybody but us

Knightly1: Fair

src/main/api/Abilities.cpp
Knightly1:
```cpp
	if (pLocalPC == nullptr)
		return false;
//snip
	return pLocalPC->GetCurrentPcProfile()->InnateSkill[nSkill] != 255;
```
?

brainiac: or

```
return pLocalPC && pLocalPC->GetCurrentPcProfile()->InnateSkill[nSkill] != 255;
```

although part of me wants to add an inline accessor to PcClinet to simplify this further like...

`pLocalPC->GetInnateSkill(nSkill) != 255`

Knightly1: I worry about the parallels in AdjustedSkill :D

include/mq/api/Abilities.h
Knightly1: From has parallels and this can include an int lookup as well as the name.  ByString would make more sense to me if it was name only...but then it would be GetAbilityIDByName.

src/main/api/Abilities.cpp
Knightly1: Fixed

src/main/api/Abilities.cpp
Knightly1: Moved these into the header and inlined them.

brainiac (approved):

PR #830, opened 2024-02-16
Remove open on arrow flag for treepanel

aquietone:
ImGuiTreeNodeFlags_OpenOnArrow  means that you may only click on the arrow to expand the tree, instead of being able to click anywhere on the text of the treenode. The entire treenode value highlights as selectable, suggesting you can click anywhere, but you can't because of the flag.

Impacts at least the MQ settings window and the autologin window.


aquietone:
oops.. removing lua eval commit from this

brainiac:
As mentioned in discord, this causes a conflictin behavior if the expandable node is also selectable. Instead, I added the flag to enable double clicks to expand.

PR #838, opened 2024-02-23
Updates examples/fish.lua

johnfking:
Updates the examples/lua.fish with:

Updated functions from function expression declaration to direct declarations
Removed use of mq.cmd.echo to output to the console, and replaced its use with print and printf
Removed use of mq.cmd.xxx commands and replaced them with the generally accepted mq.cmd('/command) syntax

### DIFF
--- a/src/plugins/lua/lua/examples/fish.lua
+++ b/src/plugins/lua/lua/examples/fish.lua
@@ -1,28 +1,28 @@
 local mq = require('mq')
-local ForceEnd = function ()
+local function ForceEnd()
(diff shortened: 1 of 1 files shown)


johnfking:
Two other things that changed....
Removed the check for Cursor() ~= 0 since the mq.TLO.Cursor() now returns nil when not holding something.
Check for an item Type() of "Fishing Pole" in the mainhand rather than a specific pole name.  This makes it more user friendly since you don't need to have a specific named pole to start fishing.

PR #855, opened 2024-04-08
feat: adds spawn constants

johnfking:


johnfking:
WIP, looking for any feedback.   Constant Names and Keys are in Pascal case.

aquietone:
Can CombatState be included too? RESTING, ACTIVE and all that.

johnfking:
> Can CombatState be included too? RESTING, ACTIVE and all that.

Yes ofc!  However, this PR was just for things that relate to the Spawn Datatype.  
Combat State is on Character I think?  That will come in the next PR once we get all the kinks out of this one as I plan to outline the character datatype next.

src/plugins/lua/bindings/lua_Constants.cpp
```diff
@@ -0,0 +1,283 @@
...
+      lua.new_enum("PlayerStates",
+         "Idle",                      0,
+         "Open",                      1,
+         "WeaponSheathed",            2,
+         "Aggressive",                4,
+         "ForcedAggressive",          8,
+         "InstrumentEquipped",        0x10,
+         "Stunned",                   0x20,
+         "PrimaryWeaponEquipped",     0x40,
+         "SecondaryWeaponEquipped",   0x80
```
brainiac, line 245: Where would these value get used?

Also, if we're going to add these values they need to be bound to actual constants instead of magic numbers.

src/plugins/lua/bindings/lua_Constants.cpp
```diff
@@ -0,0 +1,283 @@
...
+         "EmoteRude2",                 163,
+         "EmoteGrovel2",               165,
+         "EmoteBlush2",                166,
+         "EmoteCough2",                167,
+         "EmoteDuck2",                 168,
+         "EmoteShiver2",               172,
+         "EmoteTap2",                  174,
+         "EmoteGlare2",                177,
+         "EmoteStare2",                180,
+         "ZoneIn",                     254
```
brainiac, line 125: These need constants on the c++ side.

In general, if want to expose constants to lua, they need to be constants in c++ first.

brainiac:
I think we need some way to distinguish between constants that only make sense because they are exposed from TLOs.

For example, ConsiderColors are strings because Macro TLOs would only expose them as strings. But these values have actual numeric constant values, and the TLO interface just doesn't use them.

I'd like to ensure a path forward where in the future we want to expose (for exapmle) ConsiderColors as numeric constants, which would be much more efficient for comparison than strings.

src/plugins/lua/bindings/lua_Constants.cpp
johnfking: No problem.  Where would you like me to define these in c++?  Here, or in a separate header or class file?

src/plugins/lua/bindings/lua_Constants.cpp
johnfking: This was an interesting one when I wrote it.  The spawn object has a member, `PlayerState` that "returns a mask as an inttype which has the following meaning: 0=Idle 1=Open 2=WeaponSheathed 4=Aggressive 8=ForcedAggressive 0x10=InstrumentEquipped 0x20=Stunned 0x40=PrimaryWeaponEquipped 0x80=SecondaryWeaponEquipped"

So my assumption on it's use in Lua would be an easier way to check the state, such as

```lua
local mySpawn = mq.TLO.Me.Target()
if mySpawn.PlayerState() == Constants.PlayerState.SecondaryWeaponEquipped then
-- Do Stuff
end
```

johnfking: I take it back, lol.  I still think this is useful.  Updated the answer of where this could be used with an example.
```lua
local mySpawn = mq.TLO.Me.Target()
if mySpawn.PlayerState() == (Constants.PlayerState.PrimaryWeaponEquipped + Constants.PlayerState.SecondaryWeaponEquipped) then
-- Do Stuff
end
```

PR #859, opened 2024-04-16
Added some extra pulse

spik3n:
Added some extra pulse check to handle Invalid servers


spik3n:
> I'm not sure why your commit amended to the one we already merged, but I'm not going to accept this change for now. Maybe we'll do something similar later if the timing is right.

All good, this was just an ide to do it. I had another one also to mask all with a md5 hash instead for each server and also some other things to make it harder https://github.com/spik3n/macroquest/commit/364eb6b108be1bd671a5fe16ec1a9ccfc7c7dafe

PR #864, opened 2024-05-05
Added an option for imgui to ignore its off-screen detection code.

DerpleMQ2:
This helps for ISBoxer users who use mini windows that causeis ImGui to reset window positions.

### DIFF
--- a/contrib/imgui/imgui.cpp
+++ b/contrib/imgui/imgui.cpp
@@ -7169,7 +7169,8 @@ bool ImGui::Begin(const char* name, bool* p_open, ImGuiWindowFlags flags)
         {
             if (!window->ViewportOwned && viewport_rect.GetWidth() > 0 && viewport_rect.GetHeight() > 0.0f)
             {
-                ClampWindowPos(window, visibility_rect);
+                if (!GetIO().ImGuiIgnoreClampWindow)
+                    ClampWindowPos(window, visibility_rect);
             }
             else if (window->ViewportOwned && g.PlatformIO.Monitors.Size > 0)
             {
--- a/contrib/imgui/imgui.cpp
+++ b/contrib/imgui/imgui.cpp
@@ -7169,7 +7169,7 @@ bool ImGui::Begin(const char* name, bool* p_open, ImGuiWindowFlags flags)
         {
             if (!window->ViewportOwned && viewport_rect.GetWidth() > 0 && viewport_rect.GetHeight() > 0.0f)
             {
-                if (!GetIO().ImGuiIgnoreClampWindow)
+                if (!g.IO.IgnoreClampWindow)
                     ClampWindowPos(window, visibility_rect);
             }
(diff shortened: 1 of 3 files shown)


contrib/imgui/imgui.cpp
```diff
@@ -7169,7 +7169,8 @@ bool ImGui::Begin(const char* name, bool* p_open, ImGuiWindowFlags flags)
         {
             if (!window->ViewportOwned && viewport_rect.GetWidth() > 0 && viewport_rect.GetHeight() > 0.0f)
             {
-                ClampWindowPos(window, visibility_rect);
+                if (!GetIO().ImGuiIgnoreClampWindow)
```
brainiac, line 7172: You can use `g.IO` here instead of `GetIO()`

contrib/imgui/imgui.h
```diff
@@ -2305,6 +2305,8 @@ struct ImGuiIO
     ImWchar16   InputQueueSurrogate;                // For AddInputCharacterUTF16()
     ImVector<ImWchar> InputQueueCharacters;         // Queue of _characters_ input (obtained by platform backend). Fill using AddInputCharacter() helper.
 
+    bool        ImGuiIgnoreClampWindow;
```
brainiac, line 2308: this _is_ ImGui code, so `IgnoreClampWindow` would be sufficient

src/main/ImGuiManager.cpp
```diff
@@ -1203,6 +1203,18 @@ void ImGuiManager_OverlaySettings()
 	mq::imgui::HelpMarker("When enabled, MacroQuest will take over drawing the EQ Cursor when\n"
 		"it is near to or hovering over an ImGui window.");
 
+	if (ImGui::Checkbox("Ignore Window Clamping", &s_imguiIgnoreClampWindow))
+	{
+		WritePrivateProfileBool("Overlay", "ImGuiIgnoreClampWindow", s_imguiIgnoreClampWindow, mq::internal_paths::MQini);
+
+		auto& io = ImGui::GetIO();
+		if (io.ImGuiIgnoreClampWindow != s_imguiIgnoreClampWindow)
```
brainiac, line 1211: the condition is unnecessary

brainiac: You also need to set this when the context is created

brainiac (changes requested):
left some comments

src/main/ImGuiManager.cpp
DerpleMQ2: fixed.

contrib/imgui/imgui.h
DerpleMQ2: fixed.

contrib/imgui/imgui.cpp
DerpleMQ2: fixed.

DerpleMQ2:
Made the changes in the second commit

brainiac (approved):

PR #882, opened 2024-07-07
added shift to dock and keyboard nav toggles

grimmier378:
not sure how to make these settings active when mq loads. 

they do save and display saved state when you open the settings window though.


brainiac (changes requested):
You need to initialize these settings into the ImGuiIO structure in ImGuiManager_CreateContext.

If you want to follow an example, look at the references to `s_imguiIgnoreClampWindow`.

PR #884, opened 2024-07-08
added /mqlog2 command Usage: /mqlog2 <filename> <log text>

grimmier378:
This way you can debug to a specified file and not have to hunt through MacroQuest.log for what you need.


brainiac:
MQ2Log plugin already accomplishes this

grimmier378:
I was trying to avoid making people have the plugin on.
no biggie though i'll just add some checks and if its on use /mqlogcustom and if not /mqlog

PR #885, opened 2024-07-14
repositioned the notice that a plugin is being removed

hytiek:
Previously the location of this would trigger the OnUnloadPlugin before the unloading plugin was actually unloaded.

### DIFF
--- a/src/main/MQPluginHandler.cpp
+++ b/src/main/MQPluginHandler.cpp
@@ -624,9 +624,6 @@ bool UnloadPlugin(std::string_view pluginName, bool save /* = false */)
 		rec = iter->second;
 		pPlugin = rec.instance;
 
-		// Inform other plugins that this plugin is being removed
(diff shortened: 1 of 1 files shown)


Knightly1:
I'm not following, isn't the purpose of the existing order of operations to allow plugins with references to the unloading plugin to remove them before those references are invalidated?

Knightly1:
Hytiek and I discussed what he's trying to accomplish, and how to do that using the existing functionality.  Closing this out.

brainiac:
Just to add some closure to this - the reason why we do not want to do this is because in order to unregister resources from a plugin that is unloading, the plugin needs to still exist while this callback is called.

For example, if MQ2Nav was unloading, and you were doing some custom nav behaviors in your plugin, the OnUnloadPlugin callback would be called and then you would then make calls to nav to unload your custom behaviors.

if the plugin had already been unloaded at this point, it would be impossible to do.

PR #890, opened 2024-08-09
MacroQuest_Overlay.ini save and load per char

grimmier378:
when game state = ingame change to a per character ini file ServerName_CharName_Overlay.ini

When game stat is not in game switch back to using the default MacroQuest_Overlay.ini

If the character ini is missing copy the default MacroQuest_Overlay.ini to the ServerName_CharName_Overlay,ini


brainiac:
I'm not sure if you tested this, but I can tell from looking at this, that it doesn't work.

This isn't a simple problem and it has implications on existing behaviors, which is why we haven't worked on a general solution for this yet.

PR #891, opened 2024-08-12
add option to not clean links from mq.events

grimmier378:
if unchecked mq.events will not strip links from the line.

This way we preserve the item links from chat when printing the line back.

https://github.com/user-attachments/assets/0bf77243-08f5-4718-a7d1-701bd86aa9eb

### DIFF
--- a/src/plugins/lua/LuaEvent.cpp
+++ b/src/plugins/lua/LuaEvent.cpp
@@ -121,7 +121,7 @@ void LuaEventProcessor::Process(std::string_view line) const
 
 	char line_char[MAX_STRING] = { 0 };
 
-	if (line.find_first_of('\x12') != std::string::npos)
+	if (line.find_first_of('\x12') != std::string::npos && s_cleanLinks)
 	{
 		CXStr line_str(line);
 		line_str = CleanItemTags(line_str, false);
--- a/src/plugins/lua/LuaEvent.cpp
+++ b/src/plugins/lua/LuaEvent.cpp
@@ -276,7 +276,21 @@ void LuaEventProcessor::HandleBlechEvent(LuaEvent* pEvent, BLECHVALUE* pValues)
 	{
 		auto num = GetIntFromString(value->Name, 0);
 		if (num > 0) // this will skip any '*' instances for me -- it will in fact only Get valid argument positions
-			args.emplace_back(num, value->Value);
+		{
+			std::string argValue = value->Value;
+
+			// If we should not keep links, clean the argument value
+			if (!pEvent->ShouldKeepLinks() && argValue.find_first_of('\x12') != std::string::npos)
+			{
+				CXStr arg_str(argValue.c_str());
+				arg_str = CleanItemTags(arg_str, false);
+				char cleaned_arg[MAX_STRING] = { 0 };
+				StripMQChat(arg_str, cleaned_arg);
+				argValue = cleaned_arg;
+			}
+
+			args.emplace_back(num, std::move(argValue));
+		}
 		value = value->pNext;
 	}
 
--- a/src/plugins/lua/LuaEvent.cpp
+++ b/src/plugins/lua/LuaEvent.cpp
@@ -43,7 +43,7 @@ LuaEventProcessor::~LuaEventProcessor()
 	m_eventDefinitions.clear();
 }
 
-bool LuaEventProcessor::AddEvent(std::string_view name, std::string_view expression, const sol::function& function)
+bool LuaEventProcessor::AddEvent(std::string_view name, std::string_view expression, const sol::function& function, const sol::table& options)
 {
 	// the number of events will always be fairly small, and this is a manual operation.
 	// If this is deemed too slow, the event names can be memoized in a set of string_views.
@@ -56,8 +56,9 @@ bool LuaEventProcessor::AddEvent(std::string_view name, std::string_view express
 		return false;
 	}
 
-	m_eventDefinitions.push_back(std::make_unique<LuaEvent>(name, expression, function, this, *m_blech));
-	return true;
+    // Create the event with the options table
+    m_eventDefinitions.push_back(std::make_unique<LuaEvent>(name, expression, function, options, this, *m_blech));
+    return true;
 }
 
 bool LuaEventProcessor::RemoveEvent(std::string_view name)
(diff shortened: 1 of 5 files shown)


brainiac (changes requested):
This fails to account for some critical details:
- By being a user setting, this assumes that the user knows the correct setting to be used to be compatible with a running script
- It preclues the ability to run two scripts, or even events, simultaneously, where one requires stripping of chat links, and the other does not.

Note that stripping of chat links in events isn't a user preference. This is something that the event itself would need to opt into.

grimmier378:
I see your point and also agree. 

maybe making it so `mq.events(name, pattern, callback, showlink) ` where showlink is optional and defaults to false which is the current way it works, where they are always stripped?

brainiac:
I'd probably go one step further and make it be `mq.events(name, pattern, callback, options)` where options is a table.

you would call it like `mq.events(name, pattern, callback, {keep_links = True})` or something

grimmier378:
https://vimeo.com/998196547/ef502773a4?share=copy

test script used
[test_events.lua.txt](https://github.com/user-attachments/files/16600634/test_events.lua.txt)

src/plugins/lua/LuaEvent.cpp
```diff
@@ -304,13 +337,23 @@ void CALLBACK LuaEventCallback(unsigned int ID, void* pData, BLECHVALUE* pValues
...
-	const sol::function& func, LuaEventProcessor* processor, Blech& blech)
+	const sol::function& func,
+	const sol::table& options, // Accept options table
+	LuaEventProcessor* processor, Blech& blech)
 	: m_name(name)
 	, m_expression(expression)
 	, m_function(func)
 	, m_processor(processor)
 	, m_blech(blech)
+	, m_keep_links(false)
```
brainiac, line 348: `m_keepLinks`, this should be initialized in the header

src/plugins/lua/LuaEvent.cpp
```diff
@@ -304,13 +337,23 @@ void CALLBACK LuaEventCallback(unsigned int ID, void* pData, BLECHVALUE* pValues
 }
 
 LuaEvent::LuaEvent(std::string_view name, std::string_view expression,
-	const sol::function& func, LuaEventProcessor* processor, Blech& blech)
+	const sol::function& func,
+	const sol::table& options, // Accept options table
```
brainiac, line 341: redundant comment provides no value.

src/plugins/lua/MQ2Lua.cpp
```diff
@@ -1398,6 +1400,11 @@ static void DrawLuaSettings()
 		s_configNode["verboseErrors"] = s_verboseErrors;
 	}
 
+	if (ImGui::Checkbox("Clean Links from Events", &s_cleanLinks))
+	{
+		s_configNode["cleanLinks"] = s_cleanLinks;
+	}
+
```
brainiac, line 1407: You need to remove all this code from the first implementation still.

src/plugins/lua/LuaEvent.h
```diff
@@ -44,13 +45,17 @@ class LuaEvent
...
+	std::string_view GetExpression() const { return m_expression; } // Add accessor for expression
+
 private:
 	const std::string m_name;
 	const std::string m_expression;
 	const sol::function m_function;
 	LuaEventProcessor* m_processor;
 	Blech& m_blech;
 	uint32_t m_id;
+	bool m_keep_links; // Store the keep_links option
```
brainiac, line 58: Did you use AI to write all this?

src/plugins/lua/LuaEvent.cpp
```diff
@@ -304,13 +337,23 @@ void CALLBACK LuaEventCallback(unsigned int ID, void* pData, BLECHVALUE* pValues
...
 	, m_blech(blech)
+	, m_keep_links(false)
 {
+	if (options.valid())
+	{
+		if (auto keep_links_opt = options["keep_links"]; keep_links_opt.valid())
+		{
+			m_keep_links = keep_links_opt.get_or(false); // Get the value or default to false
+		}
+	}
```
brainiac, line 356: If you pass in the optional table as I described in another comment, you can do something like this. The valid checks are redundant in this case.

```
if (options.has_value())
{
    m_keepLinks = options["keep_links"].get_or_(false);
}
```

src/plugins/lua/LuaEvent.h
```diff
@@ -44,13 +45,17 @@ class LuaEvent
 	std::string_view GetName() const { return m_name; }
 	const sol::function GetFunction() const { return m_function; }
 
+	bool ShouldKeepLinks() const { return m_keep_links; } // Add accessor for keep_links
+	std::string_view GetExpression() const { return m_expression; } // Add accessor for expression
```
brainiac, line 49: ???

src/plugins/lua/LuaEvent.cpp
```diff
@@ -250,14 +242,55 @@ void LuaEventProcessor::HandleBlechEvent(LuaEvent* pEvent, BLECHVALUE* pValues)
 	std::vector<std::pair<uint32_t, std::string>> args;
 
 	std::optional<std::string> line = m_thread->GetState()["_mq_event_line"];
+	if (!line.has_value())
+		return;
```
brainiac, line 246: We shouldn't get this far if the line has no value.

src/plugins/lua/LuaEvent.cpp
```diff
@@ -250,14 +242,55 @@ void LuaEventProcessor::HandleBlechEvent(LuaEvent* pEvent, BLECHVALUE* pValues)
...
+			StripMQChat(line_str, line_char);
+		}
+		else
+		{
+			StripMQChat(line.value(), line_char);
+		}
+		// Replace the cleaned line back into the Lua state
+		line = std::string(line_char);
+		m_thread->GetState()["_mq_event_line"] = line_char;
+	}
```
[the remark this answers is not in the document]
grimmier378: I do ask copilot for help when I hit a wall, and there were a few in this one.
It loves to over comment to explain stuff to me.

brainiac:
Will pull some of this into some other link-related work that I am doing.

grimmier378:
> Will pull some of this into some other link-related work that I am doing.

copy that, on a plus note it has been working well for me in testing these last few weeks..

PR #894, opened 2024-08-30
Feat - add CustomCharacterINI to type member

Sicprofundus:
- adds CustomCharacterINI to typemember
- - ex: mq.TLO.AutoLogin.Profile.CustomCharacterIni() and ${AutoLogin.Profile.CustomCharacterINI}

### DIFF
--- a/src/plugins/autologin/MQ2AutoLogin.cpp
+++ b/src/plugins/autologin/MQ2AutoLogin.cpp
@@ -53,7 +53,8 @@ public:
 		Profile,
 		Account,
 		Class,
-		Level
+		Level,
+		CustomCharacterIni,
 	};
 
 	LoginProfileType() : MQ2Type("LoginProfile")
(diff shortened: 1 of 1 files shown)


dannuic:
Custom inis don't work

Sicprofundus:
> Custom inis don't work

well that explains the root reason why i needed/wanted to see what mq thought i was using for customini

src/plugins/autologin/MQ2AutoLogin.cpp
```diff
@@ -122,6 +124,11 @@ class LoginProfileType : public MQ2Type
 			Dest.Int = record->characterLevel;
 			Dest.Type = mq::datatypes::pIntType;
 			return true;
+		case LoginProfileMembers::CustomCharacterIni:
+			strcpy_s(DataTypeTemp, record->customClientIni ? record->customClientIni->c_str() : "No Custom INI");
```
brainiac, line 128: Wouldn't this be better if it were empty vs a magic string that everybody would need to know means "empty"?

Sicprofundus: I waffled on that a bit doing `/echo ${AutoLogin.Profile.CustomCharacterIni} and getting just an empty space seemed like it might be confusing, i figured instead of saying "none" or "null" or "" saying "No Custom INI" would be pretty clear.

that's just my rational, I'm not married to one or the other.

I also found it interesting it appears the code requires "eqclient" to be in the customini name "sic-eqclient.ini" for example, which wasn't obvious.

let me know if you would like the empty string

brainiac: Empty string seems like the universal standard when theres no value.

brainiac (approved):

PR #896, opened 2024-09-11
Added Search bar to the AA Inspection window. This will search names,…

DerpleMQ2:
… descriptions, categories, and expansions.

### DIFF
--- a/src/main/MQ2DeveloperTools.cpp
+++ b/src/main/MQ2DeveloperTools.cpp
@@ -1216,6 +1216,11 @@ public:
 
 		ImGui::Checkbox("Show Visible Only", &m_showVisable);
 
+		ImGui::Text("Search: ");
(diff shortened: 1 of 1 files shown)


src/main/MQ2DeveloperTools.cpp
```diff
@@ -1216,6 +1216,11 @@ class AltAbilityInspector : public ImGuiWindowBase
 
 		ImGui::Checkbox("Show Visible Only", &m_showVisable);
 
+		ImGui::Text("Search: ");
+	
+		static char searchText[256] = { 0 };
```
brainiac, line 1221: This is a class. Use a member variable instead of a static

DerpleMQ2:
no idea why it opened another PR but whatever I'm closing this one.

PR #900, opened 2024-10-01
Improved lua spawn functions by storing the spawn table as userdata

dannuic:


dannuic:
This strategy is not easy, implementing a better strategy in a new PR

PR #902, opened 2024-10-04
Lua spawn functions multistate

dannuic:
Create lua spawn and ground item tables in every thread's state by lazily constructing it and updating it if the script ever uses it.

### DIFF
--- a/src/plugins/lua/LuaThread.cpp
+++ b/src/plugins/lua/LuaThread.cpp
@@ -658,7 +658,6 @@ void LuaThread::AssociateTopLevelObject(const MQTopLevelObject* tlo)
 
 //============================================================================
 
-// TODO: Remove the debugging statements
 void LuaThread::InitializeSpawnTable()
 {
 	if (m_spawnTable == sol::nil)
--- a/src/plugins/lua/LuaThread.cpp
+++ b/src/plugins/lua/LuaThread.cpp
@@ -19,6 +19,7 @@
 #include "LuaImGui.h"
 #include "LuaActor.h"
 #include "bindings/lua_Bindings.h"
+#include "bindings/lua_MQBindings.h"
 
 #include <mq/Plugin.h>
 #include <luajit.h>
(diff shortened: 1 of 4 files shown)


src/plugins/lua/LuaThread.cpp
```diff
@@ -657,4 +658,80 @@ void LuaThread::AssociateTopLevelObject(const MQTopLevelObject* tlo)
...
+	{
+		sol::table spawn_table = m_globalState.create_named_table("__spawns");
+
+		if (pSpawnManager != nullptr)
+		{
+			auto spawn = pSpawnManager->FirstSpawn;
+
+			while (spawn != nullptr)
+			{
+				WriteChatf("Adding spawn %s", spawn->Name);
```
brainiac, line 674: need to remove the debug messages

src/plugins/lua/LuaThread.cpp
```diff
@@ -657,4 +658,80 @@ void LuaThread::AssociateTopLevelObject(const MQTopLevelObject* tlo)
...
+	}
+}
+
+void LuaThread::AddSpawn(eqlib::PlayerClient* spawn)
+{
+	if (m_globalState["__spawns"] != sol::nil)
+	{
+		WriteChatf("Adding spawn %s", spawn->Name);
+		m_globalState["__spawns"][spawn->SpawnID] = bindings::lua_MQTypeVar(datatypes::pSpawnType->MakeTypeVar(spawn));
+	}
```
brainiac, line 688: These functions are doing the `__spawns` lookup twice. We should save the result from the first so we can cut these down in half (they happen in bulk)

dannuic: good call -- I did that, but would it be better if I stored these tables as members on LuaThread directly? I'd have to update it when it gets set from nil, but that should work as well, no?

brainiac: yeah that would be even better if it works as expected.

dannuic: I added and tested this, with an extra guard against trying to write/access lua tables that don't exist anymore due to a finished or bad lua state

brainiac (approved):

dannuic:
merged with 22649fcb0a9af2991593999ab6c3ec3df3b911af

PR #903, opened 2024-10-10
Expose the console FontSize to lua

grimmier378:
Allows you to change the font size per console either on create or by modifying 

`console.consoleFontSize = #`

Added slider to the mqsettings window for the main console font size, this saves between sessions.

https://github.com/user-attachments/assets/fc95de88-cdb4-4055-a9bd-5e557ccfe2e1

### DIFF
--- a/src/main/MQ2ImGuiConsole.cpp
+++ b/src/main/MQ2ImGuiConsole.cpp
@@ -602,13 +602,14 @@ struct ImGuiZepConsole : public mq::imgui::ConsoleWidget, public mq::imgui::ImGu
 	std::shared_ptr<ZepConsoleSyntax> m_syntax;
 	int m_maxBufferLines = 10000;
 	bool m_autoScroll = true;
+	int m_fontSize = 13;
 	std::string m_id;
 
 	ImGuiZepConsole(std::string_view id)
 		: m_id(std::string(id))
 	{
 		SetFont(Zep::ZepTextType::UI, mq::imgui::DefaultFont, 16);
-		SetFont(Zep::ZepTextType::Text, mq::imgui::ConsoleFont, 13);
+		SetFont(Zep::ZepTextType::Text, mq::imgui::ConsoleFont, m_fontSize);  
 		SetFont(Zep::ZepTextType::Heading1, mq::imgui::DefaultFont, 28);
 		SetFont(Zep::ZepTextType::Heading2, mq::imgui::DefaultFont, 14);
 		SetFont(Zep::ZepTextType::Heading3, mq::imgui::DefaultFont, 20);
@@ -628,7 +629,7 @@ struct ImGuiZepConsole : public mq::imgui::ConsoleWidget, public mq::imgui::ImGu
 			Zep::SyntaxProvider{ "Console", Zep::tSyntaxFactory([this](Zep::ZepBuffer* pBuffer) {
 				return std::make_shared<ZepConsoleSyntax>(*pBuffer, m_theme, m_window);
 			})
-		});
+			});
 
 		m_buffer = GetEditor().InitWithText("Console", "");
 		m_buffer->SetTheme(m_theme);
@@ -856,6 +857,8 @@ struct ImGuiZepConsole : public mq::imgui::ConsoleWidget, public mq::imgui::ImGu
 
 	void Render(const ImVec2& displaySize = ImVec2()) override
 	{
+		SetFont(Zep::ZepTextType::Text, mq::imgui::ConsoleFont, m_fontSize);
+
 		if (m_deferredCursorToEnd)
 		{
 			m_deferredCursorToEnd = false;
@@ -888,6 +891,18 @@ struct ImGuiZepConsole : public mq::imgui::ConsoleWidget, public mq::imgui::ImGu
 
 		ImGuiZepEditor::Notify(message);
 	}
+
+	int GetConsoleFontSize() const override
+	{
+		return m_fontSize;
+	}
+
+	void SetConsoleFontSize(int newFontSize) override
+	{
+		m_fontSize = newFontSize;
+		SetFont(Zep::ZepTextType::Text, mq::imgui::ConsoleFont, m_fontSize);
+	}
+
 
 	bool GetAutoScroll() const override { return m_autoScroll; }
 
@@ -1777,6 +1792,19 @@ static void ConsoleSettings()
 
 		ImGui::SameLine();
 		mq::imgui::HelpMarker("Set the number of lines to keep in the scrollback buffer. Any lines above this amount will be deleted from the top of the buffer and won't be available for viewing in the console. Larger numbers here may cause performance issues like hitching or FPS slowdowns.");
+
+		ImGui::NewLine();
+
+		int newConsoleFontSize = gImGuiConsole->m_zepEditor->GetConsoleFontSize();
+		if (ImGui::SliderInt("Console Font Size", &newConsoleFontSize, 10, 30))
+		{
+			gImGuiConsole->m_zepEditor->SetConsoleFontSize(newConsoleFontSize);
+			s_consoleFontSize = newConsoleFontSize;
+			WritePrivateProfileInt("Console", "ConsoleFontSize", s_consoleFontSize, internal_paths::MQini);
+		}
+
+		ImGui::SameLine();
+		mq::imgui::HelpMarker("Adjust the font size of the console. Changes will take effect immediately and will be saved for future sessions.");
 
 		ImGui::NewLine();
 	}
--- a/src/main/MQ2ImGuiConsole.cpp
+++ b/src/main/MQ2ImGuiConsole.cpp
@@ -75,6 +75,7 @@ static bool s_consoleVisibleOnStartup = false;
 static bool s_resetConsolePosition = false;
 static bool s_setFocus = false;
 static bool s_consolePersistentCommandHistory = false;
+static int s_currentFontIndex = 2;
 
 class ImGuiConsole;
 ImGuiConsole* gImGuiConsole = nullptr;
@@ -244,6 +245,39 @@ static std::pair<std::string_view, ImU32> ParseColorTags(std::string_view line, 
 	colorStack.push_back(color);
 
 	return { { pos, (size_t)(end - pos) }, color };
+}
+
+struct FontSizeInfo
+{
+	std::string label;
+	int value;
+};
+
+std::vector<FontSizeInfo> s_fontSizeInfo = {
+	{"10", 10}, {"12", 12}, {"13", 13}, {"14", 14}, {"16", 16},
+	{"18", 18}, {"20", 20}, {"22", 22}, {"24", 24}, {"26", 26},
+	{"28", 28}, {"30", 30}
+};
+
+void FontSizePicker()
+{
+	for (int i = 0; i < s_fontSizeInfo.size(); ++i)
+	{
+		if (s_fontSizeInfo[i].value == s_consoleFontSize)
+		{
+			s_currentFontIndex = i;
+			break;
+		}
+	}
+
+	if (ImGui::Combo("Main Console Font Size", &s_currentFontIndex,
+		[](void* data, int idx, const char** out_text) {
+			*out_text = static_cast<FontSizeInfo*>(data)[idx].label.c_str();
+			return true;
+		}, s_fontSizeInfo.data(), static_cast<int>(s_fontSizeInfo.size())))
+
+		ImGui::SameLine();
+	mq::imgui::HelpMarker("Choose the font size for the main console. Changes will take effect immediately and will be saved for future sessions.");
 }
 
 //============================================================================
@@ -610,7 +644,7 @@ struct ImGuiZepConsole : public mq::imgui::ConsoleWidget, public mq::imgui::ImGu
 		: m_id(std::string(id))
 	{
 		SetFont(Zep::ZepTextType::UI, mq::imgui::DefaultFont, 16);
-		SetFont(Zep::ZepTextType::Text, mq::imgui::ConsoleFont, m_fontSize);  
+		SetFont(Zep::ZepTextType::Text, mq::imgui::ConsoleFont, m_fontSize);
 		SetFont(Zep::ZepTextType::Heading1, mq::imgui::DefaultFont, 28);
 		SetFont(Zep::ZepTextType::Heading2, mq::imgui::DefaultFont, 14);
 		SetFont(Zep::ZepTextType::Heading3, mq::imgui::DefaultFont, 20);
@@ -901,7 +935,6 @@ struct ImGuiZepConsole : public mq::imgui::ConsoleWidget, public mq::imgui::ImGu
 		m_fontSize = newFontSize;
 		SetFont(Zep::ZepTextType::Text, mq::imgui::ConsoleFont, m_fontSize);
 	}
-
 
 	bool GetAutoScroll() const override { return m_autoScroll; }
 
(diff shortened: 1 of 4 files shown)


src/main/MQ2ImGuiConsole.cpp
```diff
@@ -856,6 +857,8 @@ struct ImGuiZepConsole : public mq::imgui::ConsoleWidget, public mq::imgui::ImGu
 
 	void Render(const ImVec2& displaySize = ImVec2()) override
 	{
+		SetFont(Zep::ZepTextType::Text, mq::imgui::ConsoleFont, m_fontSize);
```
brainiac, line 860: You should not call SetFont if the font did not change

src/main/MQ2ImGuiConsole.cpp
```diff
@@ -628,7 +629,7 @@ struct ImGuiZepConsole : public mq::imgui::ConsoleWidget, public mq::imgui::ImGu
 			Zep::SyntaxProvider{ "Console", Zep::tSyntaxFactory([this](Zep::ZepBuffer* pBuffer) {
 				return std::make_shared<ZepConsoleSyntax>(*pBuffer, m_theme, m_window);
 			})
-		});
+			});
```
brainiac, line 632: extra whitespace change

brainiac (changes requested):
This doesn't compile as submitted. Also see comments.

grimmier378:
try this one.
i was able to do fresh compiles on both emu and master.

src/main/MQ2ImGuiConsole.cpp
```diff
@@ -1779,6 +1793,17 @@ static void ConsoleSettings()
 		mq::imgui::HelpMarker("Set the number of lines to keep in the scrollback buffer. Any lines above this amount will be deleted from the top of the buffer and won't be available for viewing in the console. Larger numbers here may cause performance issues like hitching or FPS slowdowns.");
 
 		ImGui::NewLine();
+
+		if (ImGui::SliderInt("Main Console Font Size", &s_consoleFontSize, 10, 30))
```
dannuic, line 1797: Changing fonts is a pretty heavy operation, this shouldn't be a slider because you want to avoid all the intermediate set operations. Something like a drop-down with explicit supported values would be much better

src/main/MQ2ImGuiConsole.cpp
```diff
@@ -1787,6 +1812,13 @@ static void ConsoleSettings()
 		s_consolePersistentCommandHistory = false;
 		WritePrivateProfileBool("MacroQuest", "ShowMacroQuestConsole", s_consoleVisibleOnStartup, mq::internal_paths::MQini);
 		WritePrivateProfileBool("Console", "PersistentCommandHistory", s_consolePersistentCommandHistory, mq::internal_paths::MQini);
+
+		s_consoleFontSize = 13;
```
brainiac, line 1816: This needs a constant

src/main/MQ2ImGuiConsole.cpp
```diff
@@ -1787,6 +1812,13 @@ static void ConsoleSettings()
...
+		WritePrivateProfileInt("Console", "ConsoleFontSize", s_consoleFontSize, internal_paths::MQini);
+		if (gImGuiConsole != nullptr)
+		{
+			gImGuiConsole->m_zepEditor->SetFont(Zep::ZepTextType::Text, mq::imgui::ConsoleFont, s_consoleFontSize);
```
brainiac, line 1820: you should be using the new function you made.

src/main/MQ2ImGuiConsole.cpp
```diff
@@ -1795,15 +1827,18 @@ void InitializeImGuiConsole()
 	s_consoleVisibleOnStartup = GetPrivateProfileBool("MacroQuest", "ShowMacroQuestConsole", false, mq::internal_paths::MQini);
 	s_consoleVisible = s_consoleVisibleOnStartup;
 	s_consolePersistentCommandHistory = GetPrivateProfileBool("Console", "PersistentCommandHistory", false, mq::internal_paths::MQini);
+	s_consoleFontSize = GetPrivateProfileInt("Console", "ConsoleFontSize", s_consoleFontSize, internal_paths::MQini);
```
brainiac, line 1830: this should be initializing with the default, not the current value.

src/main/MQ2ImGuiConsole.cpp
```diff
@@ -1795,15 +1827,18 @@ void InitializeImGuiConsole()
...
 	{
 		WritePrivateProfileBool("MacroQuest", "ShowMacroQuestConsole", s_consoleVisibleOnStartup, mq::internal_paths::MQini);
 		WritePrivateProfileBool("Console", "PersistentCommandHistory", s_consolePersistentCommandHistory, mq::internal_paths::MQini);
+		WritePrivateProfileInt("Console", "ConsoleFontSize", s_consoleFontSize, internal_paths::MQini);
 	}
 
 	AddSettingsPanel("Console", ConsoleSettings);
 
 	gImGuiConsole = new ImGuiConsole();
+	gImGuiConsole->m_zepEditor->SetFont(Zep::ZepTextType::Text, mq::imgui::ConsoleFont, s_consoleFontSize);
```
brainiac, line 1841: this is redundant with the change you already made in the constructor

src/plugins/lua/bindings/lua_ImGuiCustom.cpp
```diff
@@ -68,6 +68,7 @@ void RegisterBindings_ImGuiCustom(sol::table& ImGui)
 		"ScrollToBottom"             , &mq::imgui::ConsoleWidget::ScrollToBottom,
 		"autoScroll"                 , sol::property(&mq::imgui::ConsoleWidget::GetAutoScroll, &mq::imgui::ConsoleWidget::SetAutoScroll),
 		"maxBufferLines"             , sol::property(&mq::imgui::ConsoleWidget::GetMaxBufferLines, &mq::imgui::ConsoleWidget::SetMaxBufferLines),
+		"consoleFontSize"			 , sol::property(&mq::imgui::ConsoleWidget::GetConsoleFontSize, &mq::imgui::ConsoleWidget::SetConsoleFontSize),
```
brainiac, line 71: this is a ConsoleWidget, the "Console" prefix is redundant.

src/main/MQ2ImGuiConsole.cpp
```diff
@@ -889,6 +891,18 @@ struct ImGuiZepConsole : public mq::imgui::ConsoleWidget, public mq::imgui::ImGu
...
+	int GetConsoleFontSize() const override
+	{
+		return m_fontSize;
+	}
+
+	void SetConsoleFontSize(int newFontSize) override
+	{
+		m_fontSize = newFontSize;
+		SetFont(Zep::ZepTextType::Text, mq::imgui::ConsoleFont, m_fontSize);
+	}
```
brainiac, line 903: The console prefix in these names is redundant.

this still needs tonly be called if the font size actually changes.

src/main/MQ2ImGuiConsole.cpp
```diff
@@ -245,6 +251,18 @@ static std::pair<std::string_view, ImU32> ParseColorTags(std::string_view line,
...
+{
+	std::string label;
+	int value;
+};
+
+std::vector<FontSizeInfo> s_fontSizeInfo = {
+	{"10", 10}, {"12", 12}, {"13", 13}, {"14", 14}, {"16", 16},
+	{"18", 18}, {"20", 20}, {"22", 22}, {"24", 24}, {"26", 26},
+	{"28", 28}, {"30", 30}
+};
```
dannuic, line 264: why is this a vector of tuples and not a map?  Also, I'm assuming this is in code because the allowable values are baked in, but it seems like you're adding another maintenance point for adding a font size.

src/main/MQ2ImGuiConsole.cpp
```diff
@@ -1554,6 +1587,39 @@ static void MakeColorGradient(float frequency1, float frequency2, float frequenc
...
+void FontSizePicker()
+{
+	for (int i = 0; i < s_fontSizeInfo.size(); ++i)
+	{
+		if (s_fontSizeInfo[i].value == s_fontSize)
+		{
+			s_currentFontIndex = i;
+			break;
+		}
+	}
```
dannuic, line 1599: this is why that vector should be a map

src/main/MQ2ImGuiConsole.cpp
```diff
@@ -68,12 +68,18 @@ static const int s_userColorCommandLink = USERCOLOR_DIALOG_LINK;
...
+static const int s_defaultFontIndex = 2;
+
+static int s_fontSize = s_defaultFontSize;
 static bool s_dockspaceVisible = true;
 static bool s_consoleVisible = false;
 static bool s_consoleVisibleOnStartup = false;
 static bool s_resetConsolePosition = false;
 static bool s_setFocus = false;
 static bool s_consolePersistentCommandHistory = false;
+static int s_currentFontIndex = s_defaultFontIndex;
```
dannuic, line 82: you don't need both of these new statics, it's redundant

src/main/MQ2ImGuiConsole.cpp
dannuic: actually, looking at this a little closer, this is completely unnecessary. You _only_ need `s_fontSize`

src/main/MQ2ImGuiConsole.cpp
```diff
@@ -1554,6 +1587,39 @@ static void MakeColorGradient(float frequency1, float frequency2, float frequenc
...
+	for (int i = 0; i < s_fontSizeInfo.size(); ++i)
+	{
+		if (s_fontSizeInfo[i].value == s_fontSize)
+		{
+			s_currentFontIndex = i;
+			break;
+		}
+	}
+
+	if (ImGui::Combo("Main Console Font Size", &s_currentFontIndex,
```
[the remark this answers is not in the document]
dannuic: actually, I see why you have a vector. I forgot the combo box works entirely on index. My comments still stand (except for the map) though, you should localize the index and just set `s_fontSize` when it is selected based on the index. You don't need the extra loop at all

src/main/MQ2ImGuiConsole.cpp
```diff
@@ -1554,6 +1587,39 @@ static void MakeColorGradient(float frequency1, float frequency2, float frequenc
...
+
+		if (selectedFontSize != s_fontSize)
+		{
+			s_fontSize = selectedFontSize;
+			if (gImGuiConsole)
+				gImGuiConsole->m_zepEditor->SetFontSize(s_fontSize);
+
+			WritePrivateProfileInt("Console", "FontSize", s_fontSize, internal_paths::MQini);
+		}
+	}
```
dannuic, line 1617: This is an overly complex way to do a combo -- it's probably better to just use the standard BeginCombo/EndCombo pattern standard to imgui. I think this is reflective of how you have set up data, which should be reconsidered

brainiac: I don't think this has anything to do with how the data is stored, but BeginCombo/EndCombo is much simpler.

src/main/MQ2ImGuiConsole.cpp
brainiac: One of these values is derived from the other. We don't need both. A vector of ints would work just as well.

src/main/MQ2ImGuiConsole.cpp
```diff
@@ -68,12 +68,18 @@ static const int s_userColorCommandLink = USERCOLOR_DIALOG_LINK;
 static const int s_userColorFactionLink = USERCOLOR_FACTION_LINK;
 #endif
 
+// Some default font size
+static const int s_defaultFontSize = 13;
+static const int s_defaultFontIndex = 2;
```
brainiac, line 73: What happens if we add more font sizes? default font index is redundant if we also have a default value.

src/main/MQ2ImGuiConsole.cpp
```diff
@@ -1787,6 +1857,14 @@ static void ConsoleSettings()
 		s_consolePersistentCommandHistory = false;
 		WritePrivateProfileBool("MacroQuest", "ShowMacroQuestConsole", s_consoleVisibleOnStartup, mq::internal_paths::MQini);
 		WritePrivateProfileBool("Console", "PersistentCommandHistory", s_consolePersistentCommandHistory, mq::internal_paths::MQini);
+
+		s_fontSize = s_defaultFontSize;
+		s_currentFontIndex = s_defaultFontIndex;
```
dannuic, line 1862: reading a bit further, what is the point of `s_fontSize` at all? it always just acts as a temporary variable and the value is never actually used as-is.

dannuic: example given: here you can just write the ini as `s_defaultFontSize` and call `SetFontSize(s_defaultFontSize)`. You're already setting the index here, so just use that in the combo

src/main/MQ2ImGuiConsole.cpp
```diff
@@ -602,13 +612,14 @@ struct ImGuiZepConsole : public mq::imgui::ConsoleWidget, public mq::imgui::ImGu
...
+	int m_fontSize = s_defaultFontSize;
 	std::string m_id;
 
 	ImGuiZepConsole(std::string_view id)
 		: m_id(std::string(id))
 	{
 		SetFont(Zep::ZepTextType::UI, mq::imgui::DefaultFont, 16);
-		SetFont(Zep::ZepTextType::Text, mq::imgui::ConsoleFont, 13);
+		SetFont(Zep::ZepTextType::Text, mq::imgui::ConsoleFont, m_fontSize);
 		SetFont(Zep::ZepTextType::Heading1, mq::imgui::DefaultFont, 28);
```
dannuic, line 623: I just noticed these other font setters.... Are these actually used? If they are, they should probably be configurable as well

dannuic: Yeah, I just watched the gif again. It looks strange to increase the log size but not the title bar and menu items. Some thought needs to go into the design of this, maybe these are derived from the newly selected font size.

You should probably write a function in console that does all of that for you given a selected font size and stop calling the zep SetFont directly (you can also wrap up the "is it changed" check there given how often we asked you to check first).

As far as the design, I think it's better to derive them, but maybe it's more desirable to set them all independently? Any thoughts on that @brainiac ?

src/main/MQ2ImGuiConsole.cpp
```diff
@@ -1088,6 +1118,9 @@ class ImGuiConsole
 		int maxBufferLines = GetPrivateProfileInt("Console", "MaxBufferLines", m_zepEditor->GetMaxBufferLines(), internal_paths::MQini);
 		m_zepEditor->SetMaxBufferLines(maxBufferLines);
 		m_history = InitConsoleDatabase(m_db, current_pid);
+
+		int fontSize = GetPrivateProfileInt("Console", "FontSize", m_zepEditor->GetFontSize(), internal_paths::MQini);
+		m_zepEditor->SetFontSize(fontSize);
```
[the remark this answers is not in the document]
brainiac: It is doing that inside SetFontSize now, as I requested.

src/main/MQ2ImGuiConsole.cpp
brainiac: The other font setters aren't used by anything we've engaged the editor with so far, but it needs them or bad things could happen.

> It looks strange to increase the log size but not the title bar and menu items

It seems totally normal and fine to me. Those aren't part of the console impl so that wouldn't be the same set of changes anyhow. I don't recommend making that change, we're already lying about the font size to the editor -- what we SHOULD be doing is regenerating the font atlas, but we're not doing that here so it looks like crap when the font gets scaled up or down. It'll look worse for the smaller fonts of the UI

I think the simplest solution wins. this is simple. I don't want anything more complicated.

grimmier378: Yeah i wanted to avoid messing with the font atlas and I figured when drawing your console for lua scripts you can always adjust the imgui font size that is not in the console already with WindowFontScale. 

I have updated the lua console example and will push that into this as well.

dannuic: No wonder it looked so bad, I thought these were a set of generated font atlas sizes. Why are the values hard coded then?

grimmier378: > No wonder it looked so bad, I thought these were a set of generated font atlas sizes. Why are the values hard coded then?

no clue on that I just left them as is and changed the one I wanted to use

brainiac: These aren't actually changing the font. its stretching the font from its specified size.

brainiac:
After a lot of deliberation, I have decided not to merge this. Implementing this would make it more difficult to change/improve this in the future without breaking existing code.

Part of the reason for this is that it codifies the incorrect behavior of zooming the font, into an API that can never really properly handle this.

Second, it propogates an anti-pattern where every integration of the editor would need/want to implement a font scale setting of some kind, resulting in inconsistent or incorrect behaivor.

As a workaround, you can now zoom the font of any editor window with a couple new keybinds, ctrl+=, ctrl+- and ctrl+0.

See 7aea9f424fc4f1ccc63eea4088465e9f22f97294 for more info.

PR #904, opened 2024-10-11
Zep editor

DerpleMQ2:
Exposed the TextEditor type to lua.
This can be used to create a feature rich edit box with syntax
highlighting.

Syntax support for:
- lua
- cpp
- markdown
- tree
- toml
- cmake
- scm or scheme
- janet
- lisp
- hlsl
- gl shaders vert/frag

Code Snipet Examples:

```lua
local TextEditor       = ImGui.TextEditor.new("##TextEditor")
TextEditor:SetSyntax("lua")
TextEditor.flags = bit32.bor(ImGuiZepEditorWindowFlags.ShowLineNumbers, ImGuiZepEditorWindowFlags.ShowWrappedLineNumbers, ImGuiZepEditorWindowFlags.WrapText, ImGuiZepEditorWindowFlags.ShowIndicators)

local function RenderEditor()
    ImGui.PushFont(ImGui.ConsoleFont)
    local yPos = ImGui.GetCursorPosY()
    local footerHeight = 35
    local editHeight = (ImGui.GetWindowHeight() * .5) - yPos - footerHeight

    TextEditor:Render(ImVec2(ImGui.GetWindowWidth() * 0.98, editHeight))
    ImGui.PopFont()
end

local userText = LuaTextEditor:GetText()
```

New Enums:

- ImGuiTextEditorWindowFlags.None,
- ImGuiTextEditorWindowFlags.ShowWhiteSpace,
- ImGuiTextEditorWindowFlags.ShowCR,
- ImGuiTextEditorWindowFlags.ShowLineNumbers,
- ImGuiTextEditorWindowFlags.ShowIndicators,
- ImGuiTextEditorWindowFlags.HideScrollBar,
- ImGuiTextEditorWindowFlags.Modal,
- ImGuiTextEditorWindowFlags.WrapText,
- ImGuiTextEditorWindowFlags.HideSplitMark,
- ImGuiTextEditorWindowFlags.GridStyle,
- ImGuiTextEditorWindowFlags.ShowLineBackground,
- ImGuiTextEditorWindowFlags.ShowWrappedLineNumbers,
- ImGuiTextEditorWindowFlags.ShowAirLine,
- ImGuiTextEditorWindowFlags.HideTrailingNewline,

Lua Interface:

ImGui.TextEditor.new( id ) - Creates a new ZepEditorObject
TextEditor:Render(ImVec2(w,h)) - Renders with option display size arg
TextEditor:Clear() - Clears Text1
TextEditor:SetSyntax( syntaxName ) - sets the syntax highligher format
TextEditor:LoadContents(text) -- set the contents of the edit box, this is expensive and should be used sparingly
TextEditor.text -- Read only text property
TextEditor.fontSize = n
TextEditor.windowFlags = bit32.bor(ImGuiZepEditorWindowFlags.[...])


dannuic (commented):
why is there code reorganization _and_ changes in zep? It makes it difficult to understand what you are actually adding

src/main/ImGuiZepEditor.cpp
```diff
@@ -478,7 +490,7 @@ ZepEditor& ImGuiZepEditor::GetEditor() const
 	return *m_editor;
 }
 
-void ImGuiZepEditor::Render(const char* id, const ImVec2& displaySize)
+void ImGuiZepEditor::Render(const ImVec2& displaySize)
```
dannuic, line 493: I'm confused why this is removed, doesn't that limit functionality to never have any children?

src/main/ImGuiZepEditor.cpp
```diff
@@ -511,4 +523,75 @@ void ImGuiZepEditor::Render(const char* id, const ImVec2& displaySize)
...
+	m_buffer->Clear();
+}
+
+void ImGuiZepEditor::GetText(std::string& outBuffer) const
+{
+	outBuffer.reserve(GetTextLength());
+	auto& workingBuffer = m_buffer->GetWorkingBuffer();
+	std::copy(workingBuffer.begin(), workingBuffer.end(), std::back_inserter(outBuffer));
+	// copy 1 less character to avoid adding the null terminator into the std::string size.
+	outBuffer.pop_back();
```
dannuic, line 537: this is a confusing way to copy, I'm not sure why you want to remove the null terminator, but how is it different than just doing `std::string(buffer.begin(), buffer.end())`

src/main/ImGuiZepEditor.cpp
```diff
@@ -511,4 +523,75 @@ void ImGuiZepEditor::Render(const char* id, const ImVec2& displaySize)
...
+}
+
+void ImGuiZepEditor::SetWindowFlags(int flags) const
+{
+	m_window->SetWindowFlags(flags);
+}
+
+std::shared_ptr<mq::imgui::ImGuiZepEditor> ImGuiZepEditor::Create(std::string_view id)
+{
+	return std::make_shared<ImGuiZepEditor>(id);
```
dannuic, line 594: The constructor isn't private, why do this? It makes it seem like there's more to creating the editor than just make_shared

src/main/ImGuiZepEditor.h
```diff
@@ -38,13 +49,36 @@ class ZepEditor_ImGui;
 class ImGuiZepEditor : public Zep::IZepComponent
 {
 public:
-	ImGuiZepEditor();
+	ImGuiZepEditor(std::string_view id);
 	virtual ~ImGuiZepEditor();
 
+	MQLIB_OBJECT static std::shared_ptr<ImGuiZepEditor> Create(std::string_view id);
```
dannuic, line 55: ahh, I see now -- you're passing a shared pointer across the boundary of a dll. Is this correct?

src/main/MQ2ImGuiConsole.cpp
```diff
@@ -593,19 +599,15 @@ class ZepMode_ImGuiConsole : public Zep::ZepMode_Standard
 //----------------------------------------------------------------------------
 
 // This is the imgui container for the Zep component.
-struct ImGuiZepConsole : public mq::imgui::ConsoleWidget, public mq::imgui::ImGuiZepEditor
+struct ImGuiZepConsole : public mq::imgui::ConsoleWidget
```
dannuic, line 602: why change the inheritance chain like this? the design before was mixins, and now it's not -- what effect does that have on the current console?

DerpleMQ2:
> why is there code reorganization _and_ changes in zep? It makes it difficult to understand what you are actually adding

@brainiac requested stone refactoring. I tried to keep it minimal.

src/main/MQ2WindowInspector.cpp
```diff
@@ -1920,7 +1920,7 @@ class ImGuiWindowStringEditor
 					Zep::ZepBuffer* buffer = m_textEditor->GetEditor().InitWithText("", sv);
 				}
 
-				m_textEditor->Render("##StringEditor", ImVec2(0, ImGui::GetContentRegionAvail().y - 26));
```
dannuic, line 1923: removing this reduces the functionality of the zep editor -- it's not something that was used but keep in mind that now you need a _separate_ editor for each render whereas before you could have a text editor maintain its own state with multiple render targets

src/main/ImGuiZepEditor.h
DerpleMQ2: Yes lots of dll shananigans.

src/plugins/lua/bindings/lua_ImGuiCustom.cpp
```diff
@@ -99,6 +102,54 @@ void RegisterBindings_ImGuiCustom(sol::table& ImGui)
 			[](mq::imgui::ConsoleWidget* pThis, const ImVec4& col, std::string_view text) { pThis->AppendText(text, MQColor(col), false); }
 		)
 	);
-}
 
+	// Widgets: ZepInputBox
+	ImGui.new_usertype<mq::imgui::ImGuiZepEditor>(
+	"ZepEditor", sol::factories(&mq::imgui::ImGuiZepEditor::Create),
```
dannuic, line 108: I'm not sure we should expose the implementation detail of "zep" into lua space. We should probably just have a "TextEditor". The user doesn't need to know the tech behind it

src/main/MQ2ImGuiConsole.cpp
DerpleMQ2: Because the zep editor itself is a lua exposable object there was some overlap with the interfaces. Now console is just a specialized version of zepeditor

src/plugins/lua/bindings/lua_ImGuiCustom.cpp
```diff
@@ -99,6 +102,54 @@ void RegisterBindings_ImGuiCustom(sol::table& ImGui)
...
+		"GetFontSize", sol::overload(
+			[](mq::imgui::ImGuiZepEditor* pThis) -> int { return pThis->GetFontSize(); }
+		),
+		"SetWindowFlags", sol::overload(
+			[](mq::imgui::ImGuiZepEditor* pThis, int flags) { pThis->SetWindowFlags(flags); }
+		),
+		"GetWindowFlags", sol::overload(
+			[](mq::imgui::ImGuiZepEditor* pThis) -> int { return pThis->GetWindowFlags(); }
+		)
+	);
```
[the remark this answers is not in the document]
brainiac: @DerpleMQ2 These `sol::overloads` are unnecessary. `sol::overload` is only needed if you have more than one function with the same name (hence "overload")

src/plugins/lua/bindings/lua_ImGuiCustom.cpp
brainiac: I think I agree with this. I told derple that a lot of the naming things are TBD once I have a chance to review and see everything that is involved.

src/main/MQ2WindowInspector.cpp
brainiac: I'm not sure what you mean by "multiple render targets" in this instance. Are you saying render the same editor multiple times?

src/main/ImGuiZepEditor.cpp
brainiac: DLL boundary mess

This function can't do any allocations.

brainiac: Also it seems to be copying an extra null terminator, hence the pop_back

contrib/zep/src/window.cpp
```diff
@@ -1172,7 +1172,8 @@ void ZepWindow::DisplayLineNumbers()
             }
             else
             {
-                strNum = std::to_string(lineInfo.bufferLineNumber);
+                // Derple's school for the non-criminally insane says that Line numbers should start at 1.
```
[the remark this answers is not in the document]
brainiac: lol, but it will live in infamy in this code review.

brainiac:
> why is there code reorganization _and_ changes in zep? It makes it difficult to understand what you are actually adding

There are constexpr objects in the zep headers which make it so that it cannot be included in mq2lua without requiring to link zep.

But I'm trying to weigh the pros and cons of just crossing that bridge and linking zep. @dannuic what are your thoughts? Should we keep it completely abstracted or just pull it in?

src/main/ImGuiZepEditor.cpp
```diff
@@ -511,4 +523,75 @@ void ImGuiZepEditor::Render(const char* id, const ImVec2& displaySize)
...
+
+	return position.Move(static_cast<long>(text.length()));
+}
+
+void ImGuiZepEditor::SetText(std::string_view text, bool appendNewLine /* = false */)
+{
+	Clear();
+
+	Zep::GlyphIterator cursor = m_window->GetBufferCursor();
+	bool cursorAtEnd = m_window->IsAtBottom();
```
brainiac, line 564: this is unused.

src/main/ImGuiZepEditor.cpp
```diff
@@ -511,4 +523,75 @@ void ImGuiZepEditor::Render(const char* id, const ImVec2& displaySize)
...
+
+	std::string_view lineView = text;
+
+	std::vector<ImU32> colorStack;
+
+	InsertText(m_buffer->End(), lineView);
+
+	if (appendNewLine)
+		InsertText(m_buffer->End(), "\n");
```
brainiac, line 573: this is console specific, would prefer this only in the console.

src/main/ImGuiZepEditor.cpp
```diff
@@ -511,4 +523,75 @@ void ImGuiZepEditor::Render(const char* id, const ImVec2& displaySize)
...
+void ImGuiZepEditor::SetText(std::string_view text, bool appendNewLine /* = false */)
+{
+	Clear();
+
+	Zep::GlyphIterator cursor = m_window->GetBufferCursor();
+	bool cursorAtEnd = m_window->IsAtBottom();
+
+	std::string_view lineView = text;
+
+	std::vector<ImU32> colorStack;
```
brainiac, line 568: unused

src/main/MQ2ImGuiConsole.cpp
```diff
@@ -729,6 +720,7 @@ struct ImGuiZepConsole : public mq::imgui::ConsoleWidget, public mq::imgui::ImGu
 		case ETAG_ACHIEVEMENT:
 			color = GetColorForChatColor(s_userColorAchievementLink).ToABGR();
 			break;
+			break;
```
brainiac, line 723: extra break

src/main/MQ2Main.vcxproj.filters
```diff
@@ -711,4 +711,4 @@
       <Filter>Source Files</Filter>
     </NASM>
   </ItemGroup>
-</Project>
+</Project>
```
brainiac, line 714: unnecessary change

src/main/MQ2WindowInspector.cpp
brainiac: on one hand, we have the id stored on the widget so we don't have to provide it every frame.

on the other hand, if we want to change the id, its a bit more complicated. Do we ever have a case where we'd want to change the id of a widget we already created?

I can't really think of a practical reason why we'd want to change the id, or render twice.

dannuic: I can see a situation where we'd want to have multiple subwindows of text without wanting to instantiate an editor for each so you could track an overall editor state. if that's not important, then it's not

dannuic:
> > why is there code reorganization _and_ changes in zep? It makes it difficult to understand what you are actually adding
> 
> There are constexpr objects in the zep headers which make it so that it cannot be included in mq2lua without requiring to link zep.
> 
> But I'm trying to weigh the pros and cons of just crossing that bridge and linking zep. @dannuic what are your thoughts? Should we keep it completely abstracted or just pull it in?

My gut is to link zep

src/main/MQ2ImGuiConsole.cpp
dannuic: I would still much prefer the pattern of keeping this as separate mixin modules -- what exactly is the overlap?

src/plugins/lua/bindings/lua_ImGuiCustom.cpp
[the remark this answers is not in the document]
dannuic: I don't know if it's possible, but it would be pretty nice if you could just edit the string buffer that the editor is displaying and that calls SetText and Render for you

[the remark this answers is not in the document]
DerpleMQ2: Addressed the naming and interface complexity with my latest commit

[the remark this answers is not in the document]
DerpleMQ2: > @DerpleMQ2 These `sol::overloads` are unnecessary. `sol::overload` is only needed if you have more than one function with the same name (hence "overload")

Fixed

src/main/ImGuiZepEditor.cpp
DerpleMQ2: Fixed

src/main/ImGuiZepEditor.cpp
DerpleMQ2: Fixed.

src/main/ImGuiZepEditor.cpp
DerpleMQ2: Fixed

src/plugins/lua/bindings/lua_ImGuiCustom.cpp
DerpleMQ2: Upated

src/plugins/lua/bindings/lua_ImGuiCustom.cpp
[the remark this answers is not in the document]
DerpleMQ2: Simplified

contrib/zep/src/window.cpp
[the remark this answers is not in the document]
DerpleMQ2: Huzzah!

src/main/MQ2WindowInspector.cpp
brainiac: no, that is not viable. Replacing the text in the editor is prohibitively expensive.

brainiac: if we did subwindows that'd be supported directly by the editor.

src/plugins/lua/bindings/lua_ImGuiCustom.cpp
```diff
@@ -99,6 +102,39 @@ void RegisterBindings_ImGuiCustom(sol::table& ImGui)
...
+		"TextEditor", sol::factories(&mq::imgui::ImGuiZepEditor::Create),
+		"Render", sol::overload(
+			[](mq::imgui::ImGuiZepEditor* pThis) { pThis->Render(); },
+			[](mq::imgui::ImGuiZepEditor* pThis, ImVec2 displaySize) { pThis->Render(displaySize); }
+		),
+		"Clear", &mq::imgui::ImGuiZepEditor::Clear,
+		"SetSyntax", [](mq::imgui::ImGuiZepEditor* pThis, const char* syntaxName) { pThis->SetSyntaxProvider(syntaxName); },
+		"text", sol::property(
+			[](mq::imgui::ImGuiZepEditor* pThis) -> std::string { size_t bufferSize = pThis->GetTextLength(); std::string text; text.reserve(bufferSize); pThis->GetText(text); return text; }
+			, &mq::imgui::ImGuiZepEditor::SetText),
```
brainiac, line 117: I didn't want to make this a property because it would be easy to mis-use and set the text without realizing its side effects.

If we really want this to be a property, then make it read only, and add a "LoadContents" or similarly named function that indicates we're not just setting a string.

src/plugins/lua/bindings/lua_ImGuiCustom.cpp
```diff
@@ -99,6 +102,39 @@ void RegisterBindings_ImGuiCustom(sol::table& ImGui)
...
+		"fontSize", sol::property(&mq::imgui::ImGuiZepEditor::GetFontSize, &mq::imgui::ImGuiZepEditor::SetFontSize),
+		"flags", sol::property(&mq::imgui::ImGuiZepEditor::GetWindowFlags, &mq::imgui::ImGuiZepEditor::SetWindowFlags)
```
brainiac, line 119: aren't there other kinds of flags? Wouldn't windowFlags be more appropriate?

DerpleMQ2: Fixed.

src/plugins/lua/bindings/lua_ImGuiCustom.cpp
DerpleMQ2: Fixed.

src/main/MQ2ImGuiConsole.cpp
DerpleMQ2: ImGuiZepEditor is itself an interface to lua which is what this PR exposes.
ConsoleWidget is a virtual interface for ImGuiZepConsole to be exposed to lua

so both ImGuiZepEditor and ImGuiZepConsole are lua exposure APIs

Then ImGuiZepConsole is derived from both of them. So if it polymorph you get weird overlaps they they both have a Clear() function etc.

Long term @brainiac will have to refactor ImGuiZepConsole to not be a Console Widget and only extent ImGuiZepEditor but that is out of the scope of this PR.

src/main/MQ2Main.vcxproj.filters
DerpleMQ2: I cannot see this change locally so I am not sure how to fix it.

src/main/MQ2ImGuiConsole.cpp
dannuic: No, that's not quite it. The ConsoleWidget is the exposed piece, not anything else. This PR needs to follow the same pattern by putting a TextWidget into include/mq/imgui and exposing _that_ to plugins. Lua is a plug-in, it shouldn't be directly including stuff from main. You need to be following the same design we already had in place for console to decouple this from the main code.

DerpleMQ2: I will let @brainiac comment, but that is how I initially had it setup and he explicitly wanted to remove the additional abstraction.

dannuic: I don't agree that we should be tightly coupling _anything_ in src/main with any plugins, we've specifically worked to decouple these things. It would be better to put it into the actual plugin API, but that's a lot more work than simply following the existing pattern, and keeps the door open to more easily refactor it to be in the API. Zep _should not_ be this transparent to plugins, it's an implementation detail.

src/plugins/lua/bindings/lua_ImGuiCustom.cpp
```diff
@@ -23,6 +23,9 @@
 #include <imgui/imgui_internal.h>
 #include <imgui/misc/cpp/imgui_stdlib.h>
 #include <sol/sol.hpp>
+#include <zep/enums.h>
+
+#include "main/ImGuiZepEditor.h"
```
dannuic, line 28: We shouldn't be including anything from main in plugins, this is the indicator that the design needs to be changed

src/plugins/lua/bindings/lua_ImGuiCustom.cpp
```diff
@@ -99,6 +102,39 @@ void RegisterBindings_ImGuiCustom(sol::table& ImGui)
...
+			[](mq::imgui::ImGuiZepEditor* pThis, ImVec2 displaySize) { pThis->Render(displaySize); }
+		),
+		"Clear", &mq::imgui::ImGuiZepEditor::Clear,
+		"SetSyntax", [](mq::imgui::ImGuiZepEditor* pThis, const char* syntaxName) { pThis->SetSyntaxProvider(syntaxName); },
+		"text", sol::property(
+			[](mq::imgui::ImGuiZepEditor* pThis) -> std::string { size_t bufferSize = pThis->GetTextLength(); std::string text; text.reserve(bufferSize); pThis->GetText(text); return text; }/* read only */),
+		"LoadContents", [](mq::imgui::ImGuiZepEditor* pThis, std::string_view text) { pThis->SetText(text); },
+		"fontSize", sol::property(&mq::imgui::ImGuiZepEditor::GetFontSize, &mq::imgui::ImGuiZepEditor::SetFontSize),
+		"windowFlags", sol::property(&mq::imgui::ImGuiZepEditor::GetWindowFlags, &mq::imgui::ImGuiZepEditor::SetWindowFlags)
+	);
```
dannuic, line 120: I see that SetText is gone now, so I'm very curious what an example of use is for this widget since `text` is readonly (I think? I thought there was a sol::readonly decorator needed for that). We likely need to see an example to make sure the API is sensible before we can really comment on the actual bindings.

brainiac: I mean, i was fine with SetText/GetText but you told him to change it.

readonly and a property with only a setter are the same.

An example use case is pulling the text from the buffer to execute it as a lua script.

dannuic:
Force pushing erases comments, you should probably merge updates so we don't lose comment chains

brainiac:
> Force pushing erases comments, you should probably merge updates so we don't lose comment chains

only if the comment was on the commit and not the pull request. but I agree, force pushing not required.

src/plugins/lua/bindings/lua_ImGuiCustom.cpp
dannuic: That's not what I said, but we'll never know because of the force pushes erasing the code I commented on. I said that the implementation was leaking and I was wondering if there was a more user friendly way to implement this. I still think there's a very good way to not have to 1) detect text changes, 2) set it, and then 3) render it. #1 and #2 here are going to get messed up in userspace a lot.

dannuic: Just re-read the comment, I missed some bits because the formatting was messed up. I'm talking about this specifically:
` TextEditor:LoadContents(text) -- set the contents of the edit box, this is expensive and should be used sparingly`

I don't think it's wise to just throw this out to the user and hope it gets used correctly. What if you get a "buffer" that is actually a userdata with the string operations specified so that it naturally detects when the content changes each frame and then calls all of this behind the scenes?

brainiac: first of all, it is not an edit box. it is a text editor. It is supposed to be simple.

The buffer is not meant to be modified outside of the editor. Doing so is dangerous and requires a bunch of additional work (like managing glyph iterators and such)

dannuic: I think this is the bit that is completely unclear. When a user types in the text editor, what happens? The internal buffer gets modified? All of this API currently makes it seem like you need to manually call SetText (LoadContents -- a name change didn't change the functionality) every time the user changes text in the editor. Is that not how it works? If not, then this API needs to be made a lot clearer, because that's contrary to how most of ImGui works and users _will_ get confused

brainiac: When the user types into the editor, it changes the contents of the editor. You can serialize that out as a string (.text) or load in new contents to be parsed into the internal buffer and display in the editor

dannuic: We don't currently have a text editor to expose to lua, zep does, and it can't be used directly in lua because zep isn't available to plugins, but that's the other point -- you can't do this work _without_ creating a new imgui widget.

and you are very right that no one has written documentation, I think that we should have had at least an example of what this is supposed to look like first, which is what I said in the first place.

grimmier378: > We don't currently have a text editor to expose to lua, zep does, and it can't be used directly in lua because zep isn't available to plugins, but that's the other point -- you can't do this work _without_ creating a new imgui widget.
> 
> and you are very right that no one has written documentation, I think that we should have had at least an example of what this is supposed to look like first, which is what I said in the first place.

[texteditor.lua.txt](https://github.com/user-attachments/files/17353207/texteditor.lua.txt)

brainiac: @grimmier378 turn that into a gist. pasting a text file around sucks.

grimmier378: [<script src="https://gist.github.com/grimmier378/be326fd3f67d009cb2f7c93128c45369.js"></script>](https://gist.github.com/grimmier378/be326fd3f67d009cb2f7c93128c45369)

grimmier378:
[https://gist.github.com/grimmier378/be326fd3f67d009cb2f7c93128c45369](url)

dannuic:
> https://gist.github.com/grimmier378/5b048633aab8fdfbc5257ad48f2c9a38

This example might be a little too involved to see what's going on. It might be good in the final examples (maybe with some cleanup), but I'm looking for something focused on just what this adds, the core nominal usage, etc

brainiac:
I'm doing a major rework of how we're handling zep, which you can see in my `editor` branch. Closing this Pr for now, I'm picking up this work in that branch.

PR #907, opened 2024-10-31
Added cleanup functionality for freeing resources after shutdown

dannuic:


src/main/MQPluginHandler.cpp
```diff
@@ -637,6 +639,9 @@ bool UnloadPlugin(std::string_view pluginName, bool save /* = false */)
 	ShutdownPlugin(rec);
 
 	// Cleanup
+	// Allow other plugins to cleanup resources after plugin shutdown but before the dll is freed
+	PluginsCleanupPlugin(pPlugin->szFilename);
```
brainiac, line 643: This should go inside of ShutdownPlugin.

src/main/MQPluginHandler.cpp
```diff
@@ -508,6 +509,7 @@ int LoadPlugin(std::string_view pluginName, bool save)
 	pPlugin->LoadPlugin        = (fMQLoadPlugin)GetProcAddress(pPlugin->hModule, "OnLoadPlugin");
 	pPlugin->UnloadPlugin      = (fMQUnloadPlugin)GetProcAddress(pPlugin->hModule, "OnUnloadPlugin");
 	pPlugin->GetPluginInterface = (fMQGetPluginInterface)GetProcAddress(pPlugin->hModule, "GetPluginInterface");
+	pPlugin->CleanupPlugin     = (fMQCleanupPlugin)GetProcAddress(pPlugin->hModule, "OnCleanupPlugin");
```
brainiac, line 512: My biggest complaint is that "Cleanup" does not self-describe enough to me. I need something that reminds me that this happens AFTER the plugin is unloaded.

I would suggest OnPostUnloadPlugin or OnAfterUnloadPlugin or something similar.

We should update the MQModule interface as well. MQCommandAPI should hook into this also but I'll do that myself instead of asking you to refactor that.

brainiac (changes requested):

src/main/MQPluginHandler.cpp
dannuic: The MQModule interface did get updated for this, unless you mean to reorder the callbacks

brainiac: Sorry I meant for the MQCommandAPI part specifically.

dannuic: ahh yeah, that's probably another PR to turn that into an MQModule

brainiac (approved):

PR #914, opened 2025-01-02
MQ2TargetInfoPHs.txt

FelisMalum:
Added tOB rares to list.


FelisMalum:
Cleaning up for resubmit.

PR #932, opened 2025-05-23
Add DisplayItem window bounds persistence

cjmanca:
Adds an option to specify a position and size for item/spell information windows, which will be set each time a window opens.

Rationale: I got tired of item/spell info windows being random sizes every time I opened them. This opens them at a specific size every time. You can still move/resize them once they're open, but they'll be set back to the specified location/size the next time you open them.

### DIFF
--- a/src/plugins/itemdisplay/MQ2ItemDisplay.cpp
+++ b/src/plugins/itemdisplay/MQ2ItemDisplay.cpp
@@ -65,6 +65,13 @@ static std::pair<MQColor, std::string_view> GetEffectInfo(ItemSpellTypes effectT
 class Settings
 {
 public:
+	static constexpr inline int default_windowX = 0;
+	static constexpr inline int default_windowY = 0;
+	static constexpr inline int default_windowWidth = 400;
+	static constexpr inline int default_windowHeight = 600;
+
+	static constexpr inline bool default_persistWindowBounds = false;
+
 	static constexpr inline bool default_lootButtonsEnabled = true;
 	static constexpr inline bool default_lucyButtonEnabled = true;
 	static constexpr inline bool default_showSpellInfoOnItems = true;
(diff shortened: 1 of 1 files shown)


src/plugins/itemdisplay/MQ2ItemDisplay.cpp
```diff
@@ -1622,6 +1696,17 @@ class CItemDisplayWndOverride : public WindowOverride<CItemDisplayWndOverride, C
 		SetItem_Trampoline(pItem, flags);
 
 		s_inSetItem = false;
+
+		if (GetGameState() == GAMESTATE_INGAME && s_settings.PersistWindowBounds())
```
brainiac, line 1700: game state will always be GAMESTATE_INGAME here.

src/plugins/itemdisplay/MQ2ItemDisplay.cpp
```diff
@@ -1669,6 +1754,23 @@ class SpellDisplayHook : public CSpellDisplayWnd
 			}
 		}
 	}
+
+	DETOUR_TRAMPOLINE_DEF(void, SetSpell_Trampoline, (int SpellID, int))
+	void SetSpell_Detour(int SpellID, int flags)
+	{
+		SetSpell_Trampoline(SpellID, flags);
+
+		if (GetGameState() == GAMESTATE_INGAME && s_settings.PersistWindowBounds())
```
brainiac, line 1763: game state will always be GAMESTATE_INGAME here.

src/plugins/itemdisplay/MQ2ItemDisplay.cpp
```diff
@@ -1622,6 +1696,17 @@ class CItemDisplayWndOverride : public WindowOverride<CItemDisplayWndOverride, C
...
+		{
+			CXRect rc = GetLocation();
+			rc.left = s_settings.GetWindowX();
+			rc.top = s_settings.GetWindowY();
+			rc.right = s_settings.GetWindowX() + s_settings.GetWindowWidth();
+			rc.bottom = s_settings.GetWindowY() + s_settings.GetWindowHeight();
+
+			((CItemDisplayWnd*)this)->UpdateGeometry(rc, true, true, true, true);
```
brainiac, line 1708: This typecast is unnecessary.

brainiac (commented):
I don't like how this appears to set it up so that all the window would be stacked ontop of each other, which seems like a big issue with usability.

However, I left some comments about the current implementation.

src/plugins/itemdisplay/MQ2ItemDisplay.cpp
cjmanca: I thought so too, but I was getting random crashes when zoning, and these checks solved them.

PR #944, opened 2025-08-28
Added the ability to use "=" with /cast command.

DerpleMQ2:
i.e.: /cast =Bane

This will make sure that you do not cast Bane of Nife instead of Bane.

### DIFF
--- a/src/main/MQCommands.cpp
+++ b/src/main/MQCommands.cpp
@@ -2959,7 +2959,7 @@ void Cast(PlayerClient* pChar, const char* szLine)
 	for (int Index = 0; Index < NUM_SPELL_GEMS; Index++)
 	{
 		EQ_Spell* pSpell = GetSpellByID(GetMemorizedSpell(Index));
-		if (pSpell && ci_starts_with(pSpell->Name, szArg1))
+		if (pSpell && MaybeExactCompare(pSpell->Name, szArg1))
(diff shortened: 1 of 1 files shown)


src/main/MQCommands.cpp
```diff
@@ -2959,7 +2959,7 @@ void Cast(PlayerClient* pChar, const char* szLine)
 	for (int Index = 0; Index < NUM_SPELL_GEMS; Index++)
 	{
 		EQ_Spell* pSpell = GetSpellByID(GetMemorizedSpell(Index));
-		if (pSpell && ci_starts_with(pSpell->Name, szArg1))
+		if (pSpell && MaybeExactCompare(pSpell->Name, szArg1))
```
brainiac, line 2962: MaybeExactCompare uses ci_equals or ci_find_substr, which means this will now match anything that has Bane in its name in your example, as opposed to anything starting with Bane.

I think we need a similar function to MaybeExactCompare that is equivalent behavior but for ci_starts_with instead of ci_find_substr.

and for the record yes all of this is bonkers.

DerpleMQ2: ah I didn't catch that let me go do that.

DerpleMQ2: Addressed this in the second commit.

PR #947, opened 2025-11-04
Add GetDataTypesNames

johnfking:
Added a method to expose registered datatypes (to Lua), so that an individual has a live list of datatypes that never get out of sync with the codebase.

### DIFF
--- a/include/mq/api/MacroAPI.h
+++ b/include/mq/api/MacroAPI.h
@@ -97,7 +97,7 @@ MQLIB_API datatypes::MQ2Type* FindMQ2DataType(const char* name);
  *
  * @return Vector containing the name of each registered datatype.
  */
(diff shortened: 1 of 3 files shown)


include/mq/api/MacroAPI.h
```diff
@@ -91,6 +92,13 @@ MQLIB_API bool RemoveMQ2Type(datatypes::MQ2Type& type);
  */
 MQLIB_API datatypes::MQ2Type* FindMQ2DataType(const char* name);
 
+/**
+ * Return the list of names for all registered datatypes.
+ *
+ * @return Vector containing the name of each registered datatype.
+ */
+MQLIB_OBJECT std::vector<std::string> GetMQ2DataTypeNames();
```
brainiac: Drop the MQ2 (just use `GetDataTypeNames()`)

PR #956, opened 2026-01-06
Autologin conflicting ID fix

### LINKED ISSUE #941: Autologin: Conflicting ID when adding a character to a group
**Describe the bug**
Image

Sicprofundus:
- fixes #941
- added push/popid to DefaultCombo
- credit to https://github.com/johnfking

### DIFF
--- a/src/loader/LoaderAutoLoginImGui.cpp
+++ b/src/loader/LoaderAutoLoginImGui.cpp
@@ -203,7 +203,7 @@ template <typename Info>
 static void DefaultCombo(Info& info, const Action& select_action)
 {
 	static Info selected;
-	ImGui::PushID(&info);
+	ImGui::PushID(&selected);
 
 	static constexpr std::string_view hidden_prefix = "##";
 	const auto width = ImGui::CalcItemWidth()


src/loader/LoaderAutoLoginImGui.cpp
```diff
@@ -203,7 +203,7 @@ template <typename Info>
 static void DefaultCombo(Info& info, const Action& select_action)
 {
 	static Info selected;
-	ImGui::PushID(&info);
+	ImGui::PushID(&selected);
```
brainiac: This is actually the opposite of what you want. You want something that uniquely defines each entry

johnfking: It does.  This code is inside of a template that is used for three different forms.  The selected    static variable holds the address of each of the three.  It's just a way to uniquely identify the combo's across each of the three forms.  While you could create a unique value for each - it just seems unnecessary since you never have more than one form displayed at the same time.

brainiac: the adderss of `selected` will be the same for each version of the function, so every `info` passed for the same type of `Info` will have the same address for `selected`

brainiac: Perhaps that is good enough for this case. FYI I had a staged change to fix this (that I forgot about) that passed in an ID string like other imgui apis.

PR #957, opened 2026-01-25
option for per character/server overlay INIs

aquietone:
Add new macroquest overlay option to enable per server/character overlay INI files.

- `EnablePerCharOverlayIni` - if on, use `configDir/MacroQuset_Overlay/{Server}_{Character}_Overlay.ini` and `logDir/MacroQuest_Overlay/{Server}_{Character}_Overlay.ini` instead of the current `MacroQuest_Overlay.ini` in `configDir` and `logDir`.
- When not fully logged into a character, the original file will still be used
- Changing the option will take effect the next time ImGui context is (re)created, like zoning, logging in, /mqoverlay reload.
- Command line option is `perchar`, not sure what would really be a good name for it but perchar isn't great.

One oddity seems at character select maybe? I end up with a `MacroQuest_Overlay/Server__Overlay.log` during login, server specific but no char name overlay config.

People might realize it causes more annoyances than its worth with scripts mixing between saving locations in their own way vs depending on a global config in the existing INI vs using ### character specific entries in the global INI.

### DIFF
--- a/src/main/MacroQuest.cpp
+++ b/src/main/MacroQuest.cpp
@@ -396,8 +396,11 @@ bool InitDirectories(const std::string& iniToRead)
 		strcpy_s(gPathResources, mq::internal_paths::Resources.c_str());
 		strcpy_s(gPathEverQuest, mq::internal_paths::EverQuest.c_str());
 
-		std::filesystem::create_directories(mq::internal_paths::Config + "\\MacroQuest_Overlay", ec);
-		std::filesystem::create_directories(mq::internal_paths::Logs + "\\MacroQuest_Overlay", ec);
+		std::filesystem::create_directories(mq::internal_paths::Config + "/MacroQuest_Overlay", ec);
+		std::filesystem::create_directories(mq::internal_paths::Logs + "/MacroQuest_Overlay", ec);
+		std::string oldOverlay = fmt::format("{}/MacroQuest_Overlay.ini", mq::internal_paths::Config);
+		std::string newOverlay = fmt::format("{}/MacroQuest_Overlay/Default_Overlay.ini", mq::internal_paths::Config);
+		std::filesystem::copy_file(oldOverlay, newOverlay, std::filesystem::copy_options::skip_existing, ec);
 
 		return true;
 	}


aquietone:
Updated this with command:  
`/mqoverlay copy toonname` - will copy `configDir/MacroQuest_Overlay/<Server>_<toonname>_Overlay.ini` to the current toon ini using the current toons Server.  
`/mqoverlay copy toonname "server name"` - will copy `configDir/MacroQuest_Overlay/<server name>_<toonname>_Overlay.ini` to the current toon ini.  

Also null terminate ini file names after formatting to handle switching to characters with different length names.

src/main/ImGuiManager.cpp
```diff
@@ -1268,7 +1269,16 @@ void ImGuiManager_CreateContext()
...
 
-	fmt::format_to(ImGuiSettingsFile, "{}/MacroQuest_Overlay.ini", mq::internal_paths::Config);
+	char* result;
+	if (pLocalPC && pLocalPC->Name[0] != 0 && s_enablePerCharacterOverlayIni)
+	{
+		result = fmt::format_to(ImGuiSettingsFile, "{}/MacroQuest_Overlay/{}_{}_Overlay.ini", mq::internal_paths::Config, GetServerShortName(), pLocalPC->Name);
+	}
+	else
+	{
+		result = fmt::format_to(ImGuiSettingsFile, "{}/MacroQuest_Overlay.ini", mq::internal_paths::Config);
```
Knightly1: Given that initialisation always creates the MacroQuest_Overlay folder, I'm not sure this shouldn't be migrated to the same folder. I understand backwards compatibility, but it seems weird to keep this instead of migrating to say -- a "Default" in that folder when that folder always exists.

I'm open to thoughts on this.

src/main/ImGuiManager.cpp
```diff
@@ -1577,6 +1603,68 @@ void MQOverlayCommand(SPAWNINFO* pSpawn, char* szLine)
...
+			s_enablePerCharacterOverlayIni ? "\agOn" : "\arOff");
+
+		WritePrivateProfileBool("Overlay", "EnablePerCharacterOverlayIni", s_enablePerCharacterOverlayIni, mq::internal_paths::MQini);
+	}
+	else if (ci_equals(szArg, "copy"))
+	{
+		char szParamServer[MAX_STRING] = { 0 };
+		char szParamCharacter[MAX_STRING] = { 0 };
+		GetArg(szParamCharacter, szLine, 2);
+		GetArg(szParamServer, szLine, 3);
```
Knightly1: Part of the reason I'm not sold on the syntax command is here we go "character server" to support your defaulting to the current server but pretty much everywhere we refer to server character in that order rather than the other.

I understand why you did it, so you can have server be optional, but something feels off about doing it that way given other nomenclature. I lean toward server just being a required argument and letting the user say "current" or something like that, but not a hard stop for me or anything. Just a thought.

Regardless, if you don't have Arg2 you definitely don't have Arg3 so you can skip that by adjusting the logic here (and change how you do your show help below).

src/main/ImGuiManager.cpp
```diff
@@ -1577,6 +1603,68 @@ void MQOverlayCommand(SPAWNINFO* pSpawn, char* szLine)
...
+		}
+		else
+		{
+			if (szParamServer[0] == 0)
+			{
+				strcpy_s(szParamServer, GetServerShortName());
+			}
+
+			char sourceConfig[MAX_PATH] = { 0 };
+			char destinationConfig[MAX_PATH] = { 0 };
```
Knightly1: You don't need c-strings here nor to allocate max_path. Nothing you're doing from this point on directly requires it.

src/main/MacroQuest.cpp
```diff
@@ -395,6 +395,10 @@ bool InitDirectories(const std::string& iniToRead)
 		strcpy_s(gPathPlugins, mq::internal_paths::Plugins.c_str());
 		strcpy_s(gPathResources, mq::internal_paths::Resources.c_str());
 		strcpy_s(gPathEverQuest, mq::internal_paths::EverQuest.c_str());
+
+		std::filesystem::create_directories(mq::internal_paths::Config + "\\MacroQuest_Overlay", ec);
+		std::filesystem::create_directories(mq::internal_paths::Logs + "\\MacroQuest_Overlay", ec);
+
```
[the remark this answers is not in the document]
aquietone: The if statement above which uses `initDirectory` expects there to be macroquest INI options to let users define custom locations for any folder being initialized through that method. I couldn't really think why these would be worth allowing to point somewhere other than the config and logs folders though so didn't want to introduce keys to customize them.

I don't think MQ has any other folders it creates besides the required ones earlier in this method, which are all considered to be required directories. These ones aren't required unless a user goes and enables the option. I felt like just creating them up front here still made more sense than potentially creating them when a user enables that option.

src/main/ImGuiManager.cpp
```diff
@@ -1577,6 +1603,68 @@ void MQOverlayCommand(SPAWNINFO* pSpawn, char* szLine)
...
+		{
+			s_enablePerCharacterOverlayIni = !s_enablePerCharacterOverlayIni;
+		}
+
+		WriteChatf("Overlay INI per character is now: %s",
+			s_enablePerCharacterOverlayIni ? "\agOn" : "\arOff");
+
+		WritePrivateProfileBool("Overlay", "EnablePerCharacterOverlayIni", s_enablePerCharacterOverlayIni, mq::internal_paths::MQini);
+	}
+	else if (ci_equals(szArg, "copy"))
```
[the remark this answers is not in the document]
aquietone: I think it does match the behavior of copy layout popup when you log in a new eq character. source you select from the window and destination is implied as the character you are logged in on.

Can add some checks for when not logged into a character, agree this command doesn't make sense if not fully logged in... unless it also accepts a destination.

PR #961, opened 2026-01-31
Add plugin-registered Lua modules to MQ2Lua

Tangeleno:
Summary

  - Add a Lua module registry that plugins can register with via the MQ2Lua interface.
  - Route require() through the registry (built-ins included) with per-thread factories and dependency tracking.
  - Provide a LuaPlugin.props helper for plugin builds that need LuaJIT headers + lua51.lib.
  - Add best-effort cleanup for modules on plugin unload.

  Behavior

  - require("module") now resolves plugin-provided modules via MQ2Lua.
  - Built-in modules (mq, actors, ImGui, ImPlot, Zep) are registered through the same mechanism.
  - Scripts that required a plugin module terminate if that plugin unloads.

  Notes

  - Registration is first-come, first-served; name collisions fail.
  - Module factories are per-thread; no cross-state table sharing.

  Testing

  - Manual: verified plugin module registration, table round-trips, and error propagation.
  - Manual: verified scripts terminate on plugin unload (dependency removal).

### DIFF
--- a/src/plugins/lua/LuaModuleRegistry.cpp
+++ b/src/plugins/lua/LuaModuleRegistry.cpp
@@ -28,7 +28,7 @@ bool LuaModuleRegistry::Register(const char* name, LuaModuleFactory factory)
 	if (!name || !name[0] || !factory)
 		return false;
 
-	auto [it, inserted] = m_modules.try_emplace(name, LuaModuleEntry{ factory });
+	auto [it, inserted] = m_modules.try_emplace(name, factory);
 	return inserted;
 }
 
--- a/src/plugins/lua/LuaModuleRegistry.h
+++ b/src/plugins/lua/LuaModuleRegistry.h
@@ -25,6 +25,11 @@ using LuaModuleFactory = sol::object (*)(sol::this_state);
 
 struct LuaModuleEntry
 {
+	explicit LuaModuleEntry(LuaModuleFactory inFactory)
+		: factory(inFactory)
+	{
+	}
+
 	LuaModuleFactory factory;
 };
 


Tangeleno:
I went ahead and implemented this direction. require now recognizes both plugin. and plugin/, and we resolve using the existing canonical plugin name normalization (MQ/MQ2 prefix optional, case‑insensitive). Plugins provide a CreateLuaModule export (settled on that name to make it read more like a factory than a lifecycle hook), which MQ2Lua calls per thread to return the module object. No explicit registration/cleanup needed.

brainiac:
> I went ahead and implemented this direction. require now recognizes both plugin. and plugin/, and we resolve using the existing canonical plugin name normalization (MQ/MQ2 prefix optional, case‑insensitive). Plugins provide a CreateLuaModule export (settled on that name to make it read more like a factory than a lifecycle hook), which MQ2Lua calls per thread to return the module object. No explicit registration/cleanup needed.

image

src/plugins/lua/LuaModuleRegistry.cpp
```diff
@@ -0,0 +1,44 @@
...
+	static LuaModuleRegistry s_registry;
+	return s_registry;
+}
+
+bool LuaModuleRegistry::Register(const char* name, LuaModuleFactory factory)
+{
+	if (!name || !name[0] || !factory)
+		return false;
+
+	auto [it, inserted] = m_modules.try_emplace(name, LuaModuleEntry{ factory });
```
brainiac: minor (not worth changing unless you really want to): If you give LuaModuleEntry a constructor you could just pass factory in directly. Kinda annoying that the std doesn't have aggregate initialization for emplace yet.

Tangeleno: Easy enough change; done.

PR #967, opened 2026-02-13
Fix INI read/write. Fix Autobank for EMU

jessebevil:
RoF2 does not have access to the /autobank command. 
INI was trying to write to a path and was not assigned a file name.

### DIFF
--- a/src/plugins/autobank/MQ2AutoBank.cpp
+++ b/src/plugins/autobank/MQ2AutoBank.cpp
@@ -450,7 +450,6 @@ PLUGIN_API void InitializePlugin()
 	EzDetour(CBankWnd__WndNotification,
 		&AutoBank::BankWnd_Hook::WndNotification_Detour,
 		&AutoBank::BankWnd_Hook::WndNotification_Trampoline);
-	sprintf_s(INIFileName, MAX_PATH, "%s/AutoBank.ini", gPathConfig);
 }
 
 PLUGIN_API void ShutdownPlugin()


src/plugins/autobank/MQ2AutoBank.cpp
```diff
@@ -347,9 +348,17 @@ static void AutoBankPulse()
 	if (pProfile->GetInventorySlot(InvSlot_Cursor) != nullptr)
 	{
 		if (gbAutoInventoryInProgress)
+		{
 			DoCommandf("/autoinventory");
+		}
 		else
+		{
+#if (!IS_EXPANSION_LEVEL(EXPANSION_LEVEL_COTF))
```
brainiac:
```suggestion
#if !IS_EXPANSION_LEVEL(EXPANSION_LEVEL_COTF)
```

src/plugins/autobank/MQ2AutoBank.cpp
```diff
@@ -451,7 +450,7 @@ PLUGIN_API void InitializePlugin()
 	EzDetour(CBankWnd__WndNotification,
 		&AutoBank::BankWnd_Hook::WndNotification_Detour,
 		&AutoBank::BankWnd_Hook::WndNotification_Trampoline);
-	sprintf_s(gbAutoBankINIFile, MAX_PATH, "%s/AutoBank.ini", gPathConfig);
+	sprintf_s(INIFileName, MAX_PATH, "%s/AutoBank.ini", gPathConfig);
```
brainiac: The `INIFileName` should already have a suitable value (unless you really don't like it)

PR #976, opened 2026-02-26
AddRectFilledMultiColorRounded

grimmier378:
Adds a multicolored rect with rounded corners.

uses ImDrawFlags so you can specify which corners.

Prior draw with just AddRectFilledMultiColor
image
You can see the corners where the bars can't round 

AddRectFilledMultiColorRounded
image

You can get a nice rounded edge

### DIFF
--- a/src/plugins/lua/lua/examples/rounded_rec_multi_color.lua
+++ b/src/plugins/lua/lua/examples/rounded_rec_multi_color.lua
@@ -48,7 +48,7 @@ local function RenderWin()
         rectSize.height = ImGui.SliderFloat("Rect Height", rectSize.height, 4, 400)
 
         -- The rounding size should not be larger than half of the smaller dimension of the rectangle, otherwise it will look strange. So we calculate the larger dimension and use that to limit the rounding size.
-        local tmpLarger = rectSize.width > rectSize.height and rectSize.width or rectSize.height
+        local tmpLarger = rectSize.width < rectSize.height and rectSize.width or rectSize.height
         if roundingSize > tmpLarger / 2 then roundingSize = tmpLarger / 2 end
         roundingSize = ImGui.SliderFloat("Rounding Size", roundingSize, 0, tmpLarger / 2)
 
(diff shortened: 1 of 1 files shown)


grimmier378:
moved to src/imgui
added a demo in lua/examples

The demo isn't using ImAnim smoothing, for better gradient smoothing ImAmin is the way to go as seen in the progress bars images  above.

image

src/imgui/ImGuiDrawListExtensions.cpp
```diff
@@ -0,0 +1,132 @@
...
+ * FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
+ * IN THE SOFTWARE.
+ */
+
+#include "pch.h"
+
+ // Needs internals because we use draw_list._Data->TempBuffer and IM_FIXNORMAL2F, etc.
+#include "imgui/imgui_internal.h"
+
+#include "ImGuiDrawListExtensions.h"
```
brainiac: This should be included first before the the imgui headers

src/imgui/ImGuiDrawListExtensions.cpp
```diff
@@ -0,0 +1,132 @@
...
+
+#define IM_NORMALIZE2F_OVER_ZERO(VX,VY)     { float d2 = VX*VX + VY*VY; if (d2 > 0.0f) { float inv_len = ImRsqrt(d2); VX *= inv_len; VY *= inv_len; } } (void)0
+#define IM_FIXNORMAL2F_MAX_INVLEN2          100.0f // 500.0f (see #4053, #3366)
+#define IM_FIXNORMAL2F(VX,VY)               { float d2 = VX*VX + VY*VY; if (d2 > 0.000001f) { float inv_len2 = 1.0f / d2; if (inv_len2 > IM_FIXNORMAL2F_MAX_INVLEN2) inv_len2 = IM_FIXNORMAL2F_MAX_INVLEN2; VX *= inv_len2; VY *= inv_len2; } } (void)0
+
+namespace mq::imgui
+{
+	//tatic void AddRectFilledMultiColorRounded(const ImVec2& p_min, const ImVec2& p_max, ImU32 col_upr_left, ImU32 col_upr_right, ImU32 col_bot_right, ImU32 col_bot_left, float rounding, ImDrawFlags flags)
```
brainiac: ?

src/plugins/lua/lua/examples/rounded_rec_multi_color.lua
brainiac: lua should be indented with 4 spaces, not tabs (follow the .editorconfig rules)

src/imgui/ImGuiDrawListExtensions.h
```diff
@@ -0,0 +1,32 @@
...
+
+#pragma once
+
+#include "imgui/imgui.h"
+
+namespace mq::imgui
+{
+	// MQ extension: rounded + multicolor version of AddRectFilledMultiColor
+	IMGUI_API void AddRectFilledMultiColorRounded( ImDrawList& draw_list, const ImVec2& p_min, const ImVec2& p_max, ImU32 col_upr_left, ImU32 col_upr_right, ImU32 col_bot_right, ImU32 col_bot_left, float rounding = 0.0f, ImDrawFlags flags = 0);
+}
```
brainiac: all these files are lacking trailing newlines, as evidenced by the marker in the github display.

src/imgui/ImGuiDrawListExtensions.cpp
```diff
@@ -0,0 +1,132 @@
...
+		const int vtx_count = points_count * 2;
+		draw_list.PrimReserve(idx_count, vtx_count);
+
+		unsigned int vtx_inner_idx = draw_list._VtxCurrentIdx;
+		unsigned int vtx_outer_idx = draw_list._VtxCurrentIdx + 1;
+		for (int i = 2; i < points_count; i++)
+		{
+			draw_list._IdxWritePtr[0] = (ImDrawIdx)(vtx_inner_idx);
+			draw_list._IdxWritePtr[1] = (ImDrawIdx)(vtx_inner_idx + ((i - 1) << 1));
+			draw_list._IdxWritePtr[2] = (ImDrawIdx)(vtx_inner_idx + (i << 1));
```
brainiac: why are you using bit shifts instead of multiplication? It makes it more difficult to understand.

brainiac (changes requested):
I left some comments of things to be addressed, but if I am to accept this, I need to be able to understand it well enough to maintain it. Unfortunately, I'm having difficulty doing that right now because it comes with zero explanation or documentation.

src/imgui/ImGuiDrawListExtensions.cpp
grimmier378: I was usnig the logic of AddConvexPolyFilled 

```
// - We intentionally avoid using ImVec2 and its math operators here to reduce cost to a minimum for debug/non-inlined builds.
// - Filled shapes must always use clockwise winding order. The anti-aliasing fringe depends on it. Counter-clockwise shapes will have "inward" anti-aliasing.
```

src/imgui/ImGuiDrawListExtensions.cpp
```diff
@@ -0,0 +1,132 @@
...
+#include "pch.h"
+
+ // Needs internals because we use draw_list._Data->TempBuffer and IM_FIXNORMAL2F, etc.
+#include "imgui/imgui_internal.h"
+
+#include "ImGuiDrawListExtensions.h"
+
+#define IM_NORMALIZE2F_OVER_ZERO(VX,VY)     { float d2 = VX*VX + VY*VY; if (d2 > 0.0f) { float inv_len = ImRsqrt(d2); VX *= inv_len; VY *= inv_len; } } (void)0
+#define IM_FIXNORMAL2F_MAX_INVLEN2          100.0f // 500.0f (see #4053, #3366)
+#define IM_FIXNORMAL2F(VX,VY)               { float d2 = VX*VX + VY*VY; if (d2 > 0.000001f) { float inv_len2 = 1.0f / d2; if (inv_len2 > IM_FIXNORMAL2F_MAX_INVLEN2) inv_len2 = IM_FIXNORMAL2F_MAX_INVLEN2; VX *= inv_len2; VY *= inv_len2; } } (void)0
```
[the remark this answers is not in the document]
grimmier378: I was trying to use the same logic from ImDrawList::AddConvexPolyFilled

the macros were defined in imgui_draw.cpp

```
// On AddPolyline() and AddConvexPolyFilled() we intentionally avoid using ImVec2 and superfluous function calls to optimize debug/non-inlined builds.
// - Those macros expects l-values and need to be used as their own statement.
// - Those macros are intentionally not surrounded by the 'do {} while (0)' idiom because even that translates to runtime with debug compilers.
#define IM_NORMALIZE2F_OVER_ZERO(VX,VY)     { float d2 = VX*VX + VY*VY; if (d2 > 0.0f) { float inv_len = ImRsqrt(d2); VX *= inv_len; VY *= inv_len; } } (void)0
#define IM_FIXNORMAL2F_MAX_INVLEN2          100.0f // 500.0f (see #4053, #3366)
#define IM_FIXNORMAL2F(VX,VY)               { float d2 = VX*VX + VY*VY; if (d2 > 0.000001f) { float inv_len2 = 1.0f / d2; if (inv_len2 > IM_FIXNORMAL2F_MAX_INVLEN2) inv_len2 = IM_FIXNORMAL2F_MAX_INVLEN2; VX *= inv_len2; VY *= inv_len2; } } (void)0
```

I will make these helper functions instead.

brainiac:
Ya i'm just wondering what that would look like.

brainiac:
OK Yeah having seen what that looks like I would agree with the assessment that this is better left to the caller.

PR #977, opened 2026-02-26
MaybeExactCompare to utilize quotes

Sicprofundus:
- this would allow a "check this in quotes" to work the same way that =this item does


Sicprofundus:
i ended up adjusting my need case directly and brain gave me some context and history of why it is how it is, so I'm going to close the request

appreciate the eyeballs and consideration

PR #994, opened 2026-04-22
lua imgui: Fix BeginPopupModal and add mq.candelay()

Knightly1:
- Fixed an issue where the lua imgui binding for BeginPopupModal was not properly allowing flags
- Added mq.candelay() - returns true if the script is in a state where delay is allowed
- Cleaned up workarounds in PackageMan.lua for the flag issue
- Added a fail state in PackageMan to prevent use if it is in a state that will for sure fail. This is a prospective fix for the blocking issue.
- Added a fail state in ImguiHelper to prevent use in a similar scenario (should not reach from PackageMan, but just in case)

### DIFF
--- a/src/plugins/lua/lua/mq/PackageMan.lua
+++ b/src/plugins/lua/lua/mq/PackageMan.lua
@@ -167,7 +167,7 @@ PackageMan.Require = function(package_name, require_name, fail_message)
     local result_message = nil
     if not my_package then
         if not mq.canDelay() then
-            printf("%s :: package '%s' (require '%s') is not installed and cannot prompt to install. Check how this method is being called.", fail_message, package_name, require_name)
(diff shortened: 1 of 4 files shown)


src/plugins/lua/lua/mq/PackageMan.lua
```diff
@@ -166,6 +166,10 @@ PackageMan.Require = function(package_name, require_name, fail_message)
     local my_package = Utils.Library.Include(require_name)
     local result_message = nil
     if not my_package then
+        if not mq.canDelay() then
+            printf("%s :: package '%s' (require '%s') is not installed and cannot prompt to install. Check how this method is being called.", fail_message, package_name, require_name)
```
brainiac: hmm, I'm concerned that this wording would just produce more questions instead of answers. Maybe something along the lines of this needs to be called from the main file of the script and not from a module?

PR #995, opened 2026-04-23
Add lua folder in the right click menu

ArcReactorKC:
The Lua folder hasn't been added to the right-click menu. I thought it would make sense since we use LUA much more often than macro now.

### DIFF
--- a/src/loader/MacroQuest.cpp
+++ b/src/loader/MacroQuest.cpp
@@ -597,6 +597,7 @@ std::string internal_paths::Logs = "Logs";
 std::string internal_paths::CrashDumps = internal_paths::Logs + "\\Dumps";
 std::string internal_paths::Plugins = "Plugins";
 std::string internal_paths::Resources = "Resources";
+std::string internal_paths::Lua = "lua";
 
 bool InitializeDirectory(std::string& strPathToInit,
 	const std::string& iniKey, const std::string& iniFile,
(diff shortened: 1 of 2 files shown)


brainiac:
Adding a lua entry to the right click menu is a good idea, but this isn't correctly getting the stored path to the lua directory.

This code is creating a new config value called `"LuaPath"` which doesn't exist, so it will always get the default. It needs to read the lua directory from the lua plugin's configuration. that path isn't stored in `internal_paths`, so adding it to `internal_paths` doesn't make sense.

Knightly1:
Adding to what brainiac said, there are some knowns here that can be used. MQ2Lua is a plugin, but it's a core plugin so you always know it's going to be there, even if you don't know if it's on or not (and for adding the Explore option, you don't actually care if it's on or not).

The configuration directory for MacroQuest and the loader are shared, so you can read the configuration of the lua plugin from the configuration path, read the yaml and use that to gather the data that you need and that's a quick operation so It can be done at execution time and will always be mostly correct.

ArcReactorKC:
> Adding to what brainiac said, there are some knowns here that can be used. MQ2Lua is a plugin, but it's a core plugin so you always know it's going to be there, even if you don't know if it's on or not (and for adding the Explore option, you don't actually care if it's on or not).
> 
> The configuration directory for MacroQuest and the loader are shared, so you can read the configuration of the lua plugin from the configuration path, read the yaml and use that to gather the data that you need and that's a quick operation so It can be done at execution time and will always be mostly correct.

This is what I am actually exploring at the moment. 
I really appreciate you guys explaining this to me.

PR #1002, opened 2026-06-04
WillStackWith - Remove full buff bar check (#793)

### LINKED ISSUE #793: Stacking evaluation returns false if the characters buff bar is full
**Describe the bug**
StacksWith returns false despite being true if the character evaluating has its long buff bar full

**To Reproduce**
An a new character without Mystical Attunement AA, pop Spell vs Spell stack UI for 2 any stacking spells.
Cast buffs on that character until it reaches 16, the evaluation turns false
Drop a buff evaluation turns again true

**Expected behavior**
Stacking on Target or between spells should return its appropriate value. Any check done vs our buff bar size should only be a concern when checking stacking on self if at all.

**Additional context**
I have not tested it on a full buff bar on a max AA/level/progresion size enhanced character but ther is a high chance it works this way too

Knightly1:
- Fixes #793
- Part of the client logic for FindAffectSlot is to check if the buff bar is full and gate on that. This change temporarily disables that check when we're running a WillStackWith check, since we never care about that situation.

### DIFF
--- a/src/main/MQ2Main.h
+++ b/src/main/MQ2Main.h
@@ -326,6 +326,20 @@ MQLIB_API bool TriggeringEffectSpell(SPELL* aSpell, int i);
 MQLIB_API bool BuffStackTest(SPELL* aSpell, SPELL* bSpell, bool bIgnoreTriggeringEffects = false, bool bTriggeredEffectCheck = false);
 MQLIB_API bool WillStackWith(const EQ_Spell* testSpell, const EQ_Spell* existingSpell);
 MQLIB_API bool IsSpellTooPowerful(PlayerClient* caster, PlayerClient* target, EQ_Spell* spell);
+
+// RAII guard for stacking queries that target someone other than the local character.
+class ScopedIgnoreBuffBarFullForStacking
+{
+public:
+	ScopedIgnoreBuffBarFullForStacking();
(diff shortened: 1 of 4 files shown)


src/main/MQ2Main.h
```diff
@@ -326,6 +326,20 @@ MQLIB_API bool TriggeringEffectSpell(SPELL* aSpell, int i);
 MQLIB_API bool BuffStackTest(SPELL* aSpell, SPELL* bSpell, bool bIgnoreTriggeringEffects = false, bool bTriggeredEffectCheck = false);
 MQLIB_API bool WillStackWith(const EQ_Spell* testSpell, const EQ_Spell* existingSpell);
 MQLIB_API bool IsSpellTooPowerful(PlayerClient* caster, PlayerClient* target, EQ_Spell* spell);
+
+// RAII guard for stacking queries that target someone other than the local character.
+class ScopedIgnoreBuffBarFullForStacking
```
brainiac: is this intended to be used as a public api? If yes, we should put it into an include in mq/base instead of appending more stuff to mq2main.h

if not, we should put it in mq2internals or some other internal header.

Knightly1: Moved to MQ2Internal.h

brainiac (approved):

PR #1003, opened 2026-06-05
Guard against self-referential macro variable expansion (#45)

### LINKED ISSUE #45: Repeat of ${MacroQuest.LastCommand} will cause a Crash
In GitLab by @Knightly1 on Jan 3, 2020, 06:46

Reported by Sic

Due to the infinite recursion of ${MacroQuest.LastCommand} repeating it will cause a crash.  Steps to reproduce:
```/echo hi
/echo ${MacroQuest.LastCommand}
/echo ${MacroQuest.LastCommand}
```

There should be a recursion check on this to prevent crashes.

Knightly1:
- Fixes #45
- This does add some overhead, but I think the value of avoiding instability and footguns is worth it

### DIFF
--- a/src/main/MQDataAPI.cpp
+++ b/src/main/MQDataAPI.cpp
@@ -1237,6 +1237,50 @@ std::string HandleParseParam(std::string_view strOriginal, const bool bParseOnce
 }
 
 /**
+ * @class MacroExpansionGuard
+ *
+ * @brief Detects and breaks self-referential macro variable expansion cycles
+ *
+ * Macro expansion recurses by re-parsing the result of a variable (see ParseMacroVar).
+ * A variable whose expansion reproduces itself - directly, or by routing
+ * through other variables - therefore recurses until the stack is exhausted and the client
+ * crashes.
+ *
+ * This guard records the tokens currently being expanded. Before a variable is
+ * (re-)expanded we check whether it is already in progress. If so, we have re-entered our own
+ * expansion (a cycle that cannot terminate) so we stop instead of recursing.
+ *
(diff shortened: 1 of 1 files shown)


src/main/MQDataAPI.cpp
```diff
@@ -1236,6 +1236,50 @@ std::string HandleParseParam(std::string_view strOriginal, const bool bParseOnce
...
+ * expansion (a cycle that cannot terminate) so we stop instead of recursing.
+ *
+ * @see ParseMacroVar
+ */
+class MacroExpansionGuard
+{
+public:
+	explicit MacroExpansionGuard(std::string token)
+	{
+		s_active.push_back(std::move(token));
```
brainiac: The lifetime of the string should outlive the guard, so it should be safe to use a string_view here, right?

Knightly1: I thought about that, but I figure most tokens are short enough that it's basically free anyway and it's a bit more robust (being that this is meant to be a guard) for the guard to control the lifetime. I think it is safe to use a string_view, but I also didn't think we're gaining much by doing it in this particular case.

But I can change it if it's a sticking point (and a note to myself that I'd change IsExpanding as well).

brainiac: I think it would be better if you're worried about adding some overhead (which it sounded like you were). string_view will be much lighter in this case.

And its a constrained use case so I don't think theres any risk of misusing it.

Knightly1: Updated

brainiac (approved):

PR #1005, opened 2026-06-07
Add sounds and arguments to Profiles (#925)

### LINKED ISSUE #925: No Sound option
**Is your feature request related to a problem? Please describe.**
eqgame.exe can accept `nosou` (nosound for some reason) as a command line argument to turn off sound. We could add an option to inject that so people who want to turn off sound on all but 1 client can while only having 1 install (custom eqclient.ini doesn't work here)

**Describe the solution you'd like**
Add an option to inject `nosou` to the command line arguments for eqgame.exe

**Describe alternatives you've considered**
Multiple installs, waste of space! Custom eqclient.ini doesn't work for this

**MQ Version**
Live - applies to all

Knightly1:
- Fixes #925
- Added a Sounds toggle to profiles so that sounds can be toggled on or off on a per-profile basis
- Added an Arguments fields to pass arguments to the client on launch on a per-profile basis
- Updated the Toggle Button style to be more clear what is on/off (and keep with theme changes)
- Cleaned up some inconsistent mixed tabs/spaces in the SQL queries that were updated (most other SQL code just has tabs)

### DIFF
--- a/src/loader/ImGui.cpp
+++ b/src/loader/ImGui.cpp
@@ -172,13 +172,13 @@ bool ToggleSlider(const char* label, bool* v)
 
 	const ImU32 bg_color = ImGui::GetColorU32(style.Colors[ImGuiCol_FrameBg]);
 
-	ImU32 fg_color;
-	if (ImGui::IsItemClicked())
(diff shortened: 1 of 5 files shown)


dannuic (approved):
SQL stuff looks good

PR #1006, opened 2026-06-12
Add full custom eqclient.ini redirection (#42)

### LINKED ISSUE #42: MQ2AutoLogin - Custom Client Settings don't seem to work for Models/Stick Figures
In GitLab by @Knightly1 on Jan 3, 2020, 06:38

When using the custom eqclient.ini settings, they seem to be ignored for settings that are only available when first set out of the game (stickfigures, luclin models, etc). It would be nice if it worked for all settings in the eqclient.ini.

Knightly1:
- Fixes #42
- Adds support for full redirection of eqclient.ini to a custom location, allowing users to maintain separate configurations for different profiles
- Create a CustomClientIni.dll that we can inject early without having to change the rest of the way that MQ loads.
- Requires launching the game with MQ's loader to ensure the custom ini is used
- Setting is set from the profile so each profile can have its own custom ini path
- Allow for setting an ini that doesn't exist
- Add a clear button for clearing a previously set ini


src/loader/customini/CustomClientINI.cpp
```diff
@@ -0,0 +1,240 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ */
+
+#ifndef WIN32_LEAN_AND_MEAN
+#define WIN32_LEAN_AND_MEAN
+#endif
+
+#include <Windows.h>
```
brainiac, line 17: windows.h should be the last thing to get included so it doesn't accidentally leak stufff like its fantastic set of defines into other headers.

and the WIN32_LEAN_AND_MEAN is in our props files so we can take that out too.

src/loader/customini/CustomClientINI.vcxproj
```diff
@@ -0,0 +1,92 @@
...
+  <ItemDefinitionGroup Condition="'$(Configuration)|$(Platform)'=='Release|x64'">
+    <ClCompile>
+      <Optimization>MaxSpeed</Optimization>
+      <FunctionLevelLinking>true</FunctionLevelLinking>
+      <IntrinsicFunctions>true</IntrinsicFunctions>
+    </ClCompile>
+    <Link>
+      <EnableCOMDATFolding>true</EnableCOMDATFolding>
+      <OptimizeReferences>true</OptimizeReferences>
+    </Link>
```
brainiac: we should be able to rely on the propsheet for all our settings

src/loader/customini/CustomClientINI.cpp
```diff
@@ -0,0 +1,240 @@
...
+
+#include "mq/base/String.h"
+
+namespace
```
brainiac: difficult to tell that all of this file is in a anonymous namespace, i usually prefer just marking things static. Its a single cpp library anyways so it doesn't really matter much.

brainiac (approved):
fine with this, have a couple minor nit picky stuff for your enjoyment.

PR #1014, opened 2026-07-04
Fix buffer overflow in ${String[...].Right[N]} when N exceeds string length

CodeCompiler:
### Summary
`${String[...].Right[N]}` can overflow the fixed 2048-byte `DataTypeTemp` global buffer when `N` is larger than the string length.

### Details
In `StringMembers::Right` (positive-`N` branch), `Len` is never clamped to the source length before the copy:

```cpp
pStart = &pStart[strlen(pStart) - Len];   // strlen - Len underflows when Len > strlen
if (pStart < szString)                     // pStart is clamped back...
    pStart = szString;
memmove(DataTypeTemp, pStart, Len + 1);    // ...but Len is still huge -> writes past DataTypeTemp
```

When `Len > strlen(pStart)`, the `size_t` subtraction wraps, `pStart` is clamped back to `szString`, but `Len` stays huge — so `memmove` copies `Len + 1` bytes into the 2048-byte `DataTypeTemp`. For example `${String[abc].Right[5000]}` copies 5001 bytes into a 2048-byte buffer: global memory corruption / crash, reachable from any macro.

### Fix
Clamp `Len` to `StrLen` (already computed/in scope in this member) before computing the start pointer, mirroring how the adjacent `.Left` member already handles it. After clamping, `StrLen - Len` is non-negative and `memmove` copies exactly the trailing `Len` chars plus the null terminator.

No client-version or eqlib coupling; behavior is unchanged for valid `N`.

### DIFF
--- a/src/main/datatypes/MQ2BasicTypes.cpp
+++ b/src/main/datatypes/MQ2BasicTypes.cpp
@@ -511,10 +511,14 @@ bool MQ2StringType::GetMember(MQVarPtr VarPtr, const char* Member, char* Index, 
 			}
 			else
 			{
-				pStart = &pStart[strlen(pStart) - Len];
+				// Clamp Len to the source length. Otherwise strlen(pStart) - Len underflows
+				// (size_t wrap), and although pStart is clamped back to szString below, Len
+				// stays huge and memmove copies Len + 1 bytes past the end of the fixed
+				// DataTypeTemp buffer -> memory corruption. Mirrors the .Left handling above.
+				if (static_cast<size_t>(Len) > StrLen)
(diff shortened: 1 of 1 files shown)


Knightly1:
There's an extra guard in Left that's missing here, and these two should mirror.  I would also suggest adding to your claude.md that you don't want 4 lines of comments for a one line change unless it's explicitly not evident what the change is regarding.

CodeCompiler:
Good catch - added the `Len > MAX_STRING - 1` clamp that .Left applies up front, so .Right now caps against an oversized source too, not just an oversized N. The two members are structurally identical now, so I dropped the comment.

PR #1017, opened 2026-07-05
Return NULL for empty XTarget slots instead of a stale name

### LINKED ISSUE #918: XTarget stores the name of mobs in empty slots instead of returning nil
XTarget still shows the last thing in the slot. When you check .ID etc it's nil.

Per Brainiac: it appears that it just returns whatever is stored as the name when you request the name without checking anything else

**To Reproduce**
mq.TLO.Me.XTarget(1)() . Use on an empty slot that had a mob in it before

**Expected behavior**
Return nil for empty slots

Image

**MQ Version**
Tested on emu

CodeCompiler:
Fixes #918

### Bug

When an XTarget slot empties, the client clears `SpawnID` and sets `XTargetSlotStatus = eXTSlotEmpty` — but it does **not** clear the `Name` buffer. The xtarget datatype reads `xts->Name` unconditionally, so `${Me.XTarget[1]}` / `mq.TLO.Me.XTarget(1)()` keeps returning the previous occupant's name instead of NULL (as reported and diagnosed in the issue).

### Fix

Guard both readers with the same `XTargetSlotStatus != eXTSlotEmpty` test the codebase already uses as the canonical slot-emptiness check (`MQ2CharacterType.cpp`, `MQ2Utilities.cpp` GetAuraSpawnName):

- **`Name` member** — empty slot now takes the existing `"NULL"` fallback branch.
- **`ToString`** — empty slot now writes `"NULL"` instead of the stale name. Per review, ToString keeps its always-return-true contract.

`eXTSlotDifferentZone` slots (e.g. a group member in another zone: valid name, `SpawnID` 0) intentionally still return the name — only `eXTSlotEmpty` is excluded.

### DIFF
--- a/src/main/datatypes/MQ2XTargetType.cpp
+++ b/src/main/datatypes/MQ2XTargetType.cpp
@@ -103,11 +103,13 @@ bool MQ2XTargetType::ToString(MQVarPtr VarPtr, char* Destination)
 		&& xts->XTargetSlotStatus != eXTSlotEmpty)
 	{
 		strcpy_s(Destination, MAX_STRING, xts->Name);
-		return true;
+	}
+	else
+	{
+		strcpy_s(Destination, MAX_STRING, "NULL");
 	}
 
-	strcpy_s(Destination, MAX_STRING, "NULL");
-	return false;
+	return true;
 }
 
 bool MQ2XTargetType::Downcast(const MQVarPtr& fromVar, MQVarPtr& toVar, MQ2Type* toType)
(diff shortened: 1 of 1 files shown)


src/main/datatypes/MQ2XTargetType.cpp
```diff
@@ -99,16 +99,15 @@ bool MQ2XTargetType::ToString(MQVarPtr VarPtr, char* Destination)
...
-	}
-	else
-	{
-		strcpy_s(Destination, MAX_STRING, "NULL");
+		return true;
 	}
 
-	return true;
+	strcpy_s(Destination, MAX_STRING, "NULL");
+	return false;
```
brainiac: It is expected that an override of MQ2Type::ToString always returns true.

CodeCompiler:
Updated - ToString now keeps the always-return-true contract: empty slots write NULL and return true, with the eXTSlotEmpty guard retained. PR description updated to match.

PR #1018, opened 2026-07-05
Spawn.Invis[type] no longer reports the local player's invis state for other spawns

### LINKED ISSUE #387: Spawn Invis[#] checks local PC instead of Spawn
In GitLab by @aquietone on Sep 21, 2021, 09:27

```
${Spawn[=someoneelse].Invis}
${Spawn[=someoneelse].Invis[1]}
${Spawn[=someoneelse].Invis[2]}
${Spawn[=myself].Invis[2]}
```
1. Returns correct true/false depending whether someoneelse is invis
2. Returns true/false based on whether myself is invis
3. Returns true/false based on whether myself is ivu
4. Returns true/false based on whether myself is ivu

Not sure what all information the client has available, whether it can actually determine the different kinds of invis on external players, but either way it seems misleading that it always returns values about myself even if I'm looking at some other spawn.

Some relevant discussion here: https://discord.com/channels/511690098136580097/840375268685119499/889938182500671508

From dannuic/knightly: for 1,2,3,4 it checks pLocalPC instead of pSpawn.

CodeCompiler:
Fixes #387

### Bug

In the `SpawnMembers::Invis` handler, the unindexed and `[ANY]` cases correctly read `pSpawn->HideMode`, but the type-specific branches query the **local character** regardless of which spawn the type var refers to:

```cpp
case InvisModes::Regular:
    Dest.Set(pLocalPC->CalculateInvisLevel(eAll) != 0);   // local player, not pSpawn
```

So `${Spawn[x].Invis[1]}` / `[UNDEAD]` / `[ANIMAL]` / `[SOS]` report the local player's invisibility state for every spawn (exactly as the issue describes: "for 1,2,3,4 it checks pLocalPC instead of pSpawn").

### Fix

A per-spawn equivalent doesn't exist — `CalculateInvisLevel` is a `CharacterZoneClient` method and only the local player has one; other spawns (`PlayerClient`) expose only the binary `int HideMode` with no invis-type breakdown. So: keep the detailed logic only when `pSpawn == pLocalPlayer` (the identity comparison already used later in this function) and fall back to `HideMode` for any other spawn — the same value the unindexed `Invis` member already returns.

Behavior: `${Me.Invis[...]}` unchanged; `${Spawn[=other].Invis}` / `[ANY]` unchanged; `${Spawn[=other].Invis[1..4]}` now reports whether **that spawn** is invis at all (the best the client data allows) instead of the local player's state. Invalid indices still return FALSE (guard sits after index validation).

### DIFF
--- a/src/main/datatypes/MQ2SpawnType.cpp
+++ b/src/main/datatypes/MQ2SpawnType.cpp
@@ -617,6 +617,14 @@ bool MQ2SpawnType::GetMember(SPAWNINFO* pSpawn, const char* Member, char* Index,
 				return true;
 		}
 
+		// The type of invisibility can only be determined for the local player.
+		// For other spawns, the client only knows whether they are invis at all.
+		if (pSpawn != pLocalPlayer)
+		{
+			Dest.Set(pSpawn->HideMode != 0);
+			return true;
+		}
+
 		switch (mode)
 		{
 		case InvisModes::Any:
--- a/src/main/datatypes/MQ2SpawnType.cpp
+++ b/src/main/datatypes/MQ2SpawnType.cpp
@@ -576,82 +576,13 @@ bool MQ2SpawnType::GetMember(SPAWNINFO* pSpawn, const char* Member, char* Index,
 		Dest.Type = pBoolType;
 		return true;
 
-	case SpawnMembers::Invis: {
+	case SpawnMembers::Invis:
+		// The type of invisibility can only be determined for the local player - see
+		// Invis on the character type. For spawns, the client only knows whether they
+		// are invis at all.
+		Dest.Set(pSpawn->HideMode != 0);
 		Dest.Type = pBoolType;
-		Dest.Set(false);
-
-		if (!Index[0])
-		{
-			Dest.Set(pSpawn->HideMode != 0);
-			return true;
-		}
-
-		enum class InvisModes {
-			Any = 0,
-			Regular = 1,
-			Undead = 2,
-			Animal = 3,
-			SoS = 4,
-		};
-		InvisModes mode = InvisModes::Any;
-
-		if (IsNumber(Index))
-		{
-			mode = static_cast<InvisModes>(GetIntFromString(Index, -1));
-			if (mode < InvisModes::Any || mode > InvisModes::SoS)
-				return true;
-		}
-		else
-		{
-			if (ci_equals(Index, "ANY"))
-				mode = InvisModes::Any;
-			else if (ci_equals(Index, "NORMAL"))
-				mode = InvisModes::Regular;
-			else if (ci_equals(Index, "UNDEAD"))
-				mode = InvisModes::Undead;
-			else if (ci_equals(Index, "ANIMAL"))
-				mode = InvisModes::Animal;
-			else if (ci_equals(Index, "SOS"))
-				mode = InvisModes::SoS;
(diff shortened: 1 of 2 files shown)


src/main/datatypes/MQ2SpawnType.cpp
```diff
@@ -617,6 +617,14 @@ bool MQ2SpawnType::GetMember(SPAWNINFO* pSpawn, const char* Member, char* Index,
 				return true;
 		}
 
+		// The type of invisibility can only be determined for the local player.
+		// For other spawns, the client only knows whether they are invis at all.
+		if (pSpawn != pLocalPlayer)
+		{
+			Dest.Set(pSpawn->HideMode != 0);
+			return true;
+		}
```
brainiac: one could argue that this pLocalPC code doesn't belong in SpawnType at all and should be moved to CharacterType

CodeCompiler:
Agreed — the type-specific branches only make sense for the local player, so `CharacterType` is arguably their real home. Concretely that would look like: an `Invis` case in `MQ2CharacterType::GetMember` owning the mode parsing + `pLocalPC` logic (`CalculateInvisLevel` / SoS check), and `SpawnType::Invis` reduced to `HideMode` for every spawn and any index. One behavioral note with the move: `${Spawn[id-of-myself].Invis[UNDEAD]}` (going through the spawn TLO rather than `Me`) would then report plain `HideMode` too, since only the character type would carry the detailed logic — which seems consistent with the type split.

Happy to push that restructure into this PR, or keep this one as the minimal behavior fix and do the move as a follow-up — whichever you prefer.

CodeCompiler:
And Free Fable til the 7th so Swarm Fixes takes a couple minutes now ...

CodeCompiler:
Done - pushed the move. CharacterType now owns the mode parsing + pLocalPC logic in a new CharacterMembers::Invis, and SpawnMembers::Invis is just the spawn's HideMode for any index. Me.Invis[...] behaves exactly as before (served by the character type instead of the spawn fallthrough).

brainiac:
just fyi, this will need to be tested

CodeCompiler:
Well let me know I am finishing pulling all the offsets / structures / pretty much using Fable to completely disassemble eqgame.exe and build out a full auto compiler and updater system which has been tested working for last patch. I will update the platform with Fable tonight also to make sure its just a absolute monster.. But while we have free Claude Swarm Fable for the next 72 hours I am going to use it non stop.

CodeCompiler:
So anything else you might need lmk we can test / patch anything... LITERALLY anything... :D

brainiac:
> Well let me know I am finishing pulling all the offsets / structures / pretty much using Fable to completely disassemble eqgame.exe and build out a full auto compiler and updater system which has been tested working for last patch. I will update the platform with Fable tonight also to make sure its just a absolute monster.. But while we have free Claude Swarm Fable for the next 72 hours I am going to use it non stop.

you're not the only one using claude to try to update offsets/structures. unfortunately even the best claudes still don't understand the majority of this and its still most accurately done by hand.

(compare the current live and test releases of redguides if you wanna compare/contrast the accuracy)

brainiac:
i'm out of time for now, will follow up on these later

CodeCompiler:
Pretty sure im banned on RG :D or I would.

PR #1020, opened 2026-07-05
Window List: only treat a comma as a column selector when numeric

### LINKED ISSUE #425: Listbox selection with comma
In GitLab by @jandegithub on Dec 3, 2021, 22:10

a comma causes finding list position to fail, anything with a comma doesn't get found, thought that quoting it would imply that the comma should not refer to the column

e.g

`/varset pos ${Window[RewardSelectionWnd/RewardPageTabWindow].Tab[${tab}].Child[RewardSelectionOptionList].List["=${itemname}"]} `

e.g itemname is one of the following
-  Xril'et Log, Page 9
-  Xril'et Log, Page 15
-  Goblin Diary, Page 1

Problem was seen in the Reward window while claiming collection items from a dispenser.

CodeCompiler:
Fixes #425

### Bug

`${Window[TradeskillWnd].Child[RecipeList].List[=Xril'et Log, Page 9]}` can never match. The `List` index parser unconditionally truncates at the **first** comma and parses the remainder as a column number:

```cpp
if (char* pComma = strchr(Index, ','))
{
    nColumn = GetIntFromString(&pComma[1], nColumn) - 1;  // " Page 9" -> 0
    ...
    *pComma = '\0';                                       // needle becomes "=Xril'et Log"
}
```

Quoting can't work around it: `ParseMQ2DataPortion` strips double quotes from bracket indices before the type member ever sees them, so quoted and unquoted commas arrive identically.

### Fix

The `strrchr` approach suggested in the issue thread: take the **last** comma, and only consume the suffix as a column selector when it actually parses as a number.

- `List[text,2]` / `List[text, 2]` — byte-identical behavior, including the historical clamp of `,0`/negative to column 0.
- `List[text, with comma]` — comma now stays part of the search text (the bug).
- `List[text, with comma,2]` — still selects column 2.

Item text genuinely ending in `", <number>"` remains ambiguous with the column syntax, same as today — unavoidable without a syntax break.

### DIFF
--- a/src/main/datatypes/MQ2WindowType.cpp
+++ b/src/main/datatypes/MQ2WindowType.cpp
@@ -565,13 +565,21 @@ bool MQ2WindowType::GetMember(MQVarPtr VarPtr, const char* Member, char* Index, 
 		}
 
 		int nColumn = 0;
-		if (char* pComma = strchr(Index, ','))
+
+		// The index can optionally select a column with a ", #" suffix. Only treat the
+		// comma as a column separator when the text following it is a number, because
+		// list item text can itself contain commas (#425).
+		if (char* pComma = strrchr(Index, ','))
 		{
-			nColumn = GetIntFromString(&pComma[1], nColumn) - 1;
-			if (nColumn < 0) nColumn = 0;
-			if (nColumn >= pListWnd->Columns.GetCount())
-				return false;
-			*pComma = '\0';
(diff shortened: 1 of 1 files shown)


brainiac:
Sure, why not.

CodeCompiler:
;) I will patch the others here in a second also.

brainiac:
thanks for spending your tokens on me.

CodeCompiler:
Well helps I have unlimited...

PR #1021, opened 2026-07-05
Accept s/m/ms suffixes on the /timed delay argument

### LINKED ISSUE #417: /timed to accept "s" and "m". example: /timed 1s (for 1 second) and /timed 1m (for 1 minute)
In GitLab by @sicprofundus on Nov 27, 2021, 20:55

It would be nice that /timed accepted an "s" and/or "m" on the numbers, it would help with user issues that get tripped up on deciseconds

example: /timed 1s (for 1 second) and /timed 1m (for 1 minute)

CodeCompiler:
Fixes #417 (approved by @brainiac in the issue comments: "sure, why not. sounds like a good improvement")

### Change

`/timed` only understood raw deciseconds — `GetIntFromString` parses the leading digits, so `/timed 1s <command>` silently ran after 1 decisecond. This mirrors the exact suffix logic `/delay` already has (`Delay()` in `MQ2MacroCommands.cpp`): same base unit (deciseconds), same multipliers.

- `/timed 30 /echo hi` — unchanged (30 deciseconds)
- `/timed 1s /echo hi` — 1 second
- `/timed 2m /echo hi` — 2 minutes
- `/timed 500ms /echo hi` — 500 ms (comes with the `/delay` pattern for consistency; integer-truncated to deciseconds, same as `/delay`)

The usage string is updated to advertise the suffixes. `szArg[len - 1]` is safe: parsing sits behind the existing non-empty argument early-returns.

### DIFF
--- a/src/main/MQCommands.cpp
+++ b/src/main/MQCommands.cpp
@@ -4055,21 +4055,28 @@ void DoTimedCmd(PlayerClient* pChar, const char* szLine)
 	if (!szRest[0])
 		return;
 
-	int delay = GetIntFromString(szArg, 0);
-	size_t len = strlen(szArg);
+	// Parse the number, then interpret the unit from the exact remainder of the string.
+	// No suffix means deciseconds, matching the historical behavior.
+	int value = 0;
+	const std::from_chars_result result = std::from_chars(szArg, szArg + strlen(szArg), value);
+	const std::string_view suffix = result.ptr;
 
-	// Measured in deciseconds...
-	if (::tolower(szArg[len - 1]) == 'm')
-		delay *= 600;
-	else if (::tolower(szArg[len - 1]) == 's')
+	int delayMs;
+	if (suffix.empty())
+		delayMs = value * 100;
+	else if (ci_equals(suffix, "ms"))
+		delayMs = value;
+	else if (ci_equals(suffix, "s"))
+		delayMs = value * 1000;
+	else if (ci_equals(suffix, "m"))
+		delayMs = value * 60000;
+	else
 	{
-		if (len > 2 && ::tolower(szArg[len - 2]) == 'm')
-			delay /= 100;
-		else
-			delay *= 10;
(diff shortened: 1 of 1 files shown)


src/main/MQCommands.cpp
```diff
@@ -4055,7 +4055,21 @@ void DoTimedCmd(PlayerClient* pChar, const char* szLine)
...
+	// Measured in deciseconds...
+	if (::tolower(szArg[len - 1]) == 'm')
+		delay *= 600;
+	else if (::tolower(szArg[len - 1]) == 's')
+	{
+		if (len > 2 && ::tolower(szArg[len - 2]) == 'm')
+			delay /= 100;
+		else
+			delay *= 10;
+	}
```
brainiac: this is a bit awkward since it is just grabbing at characters from the string, not necessarily the suffix of the number (i.e. gibberish like 1000sm would be interpreted as minutes)

GetIntFromString is just a wrapper around std::from_chars, I think it would be an improvement here to use std::from_chars directly and then use a string_view of the remainder returned by from_chars with mq::ci_equals. Also, you could calculate the number in milliseconds and then divide by 10 in the call to TimedCommand to improve the clarity of this code.

CodeCompiler: Ill update it to do that.

CodeCompiler:
Reworked as suggested — `std::from_chars` directly, with `mq::ci_equals` on a `string_view` of the exact remainder, so `1000sm` (or any other trailing gibberish) is now a syntax error instead of being misread as minutes. No suffix keeps the historical deciseconds meaning.

One small deviation from the comment: I compute the delay in milliseconds and pass it straight through, without a division — `TimedCommand`'s parameter is milliseconds (`msDelay` is added directly to `MQGetTickCount64()`), and the old `deciseconds * 100` was the ds→ms conversion. Each suffix branch now states its unit conversion plainly. Happy to adjust if you'd prefer different semantics for the no-suffix case.

PR #1022, opened 2026-07-05
Fix crash when Window.Text exceeds the data buffer

### LINKED ISSUE #300: Trying to read a large amount of window text causes CTD
Enter bazaar and execute /bazaar trader to open trader window.

Execute /echo ${Window[BazaarWnd].Child[BZW_BazaarText].Text}

MQ2 window shows the expected text

Then enter trader mode, causing a lot of text to appear in the window

Execute /echo ${Window[BazaarWnd].Child[BZW_BazaarText].Text}

Result: crash to desktop

CodeCompiler:
Fixes #300

### Bug

`${Window[BazaarWnd].Child[BZW_BazaarText].Text}` (or any window whose text exceeds 2048 characters) crashes to desktop. `DataTypeTemp` is a fixed `SGlobalBuffer` of 2048 bytes, and `strcpy_s` with a longer source invokes the CRT invalid-parameter handler, which terminates the process — a guaranteed CTD whenever `GetSTMLText()`/`GetWindowText()`/`TabText` returns ≥ 2048 chars (bazaar trader mode easily does).

### Fix

Use the existing `strncpy_s(SGlobalBuffer&, const char*, _TRUNCATE)` helper in all three branches of `WindowMembers::Text` — the same pattern already used by `GetCXStr` (`MQ2Inlines.h`) and `EmuExtensions.cpp`. Short text is bit-identical to before; long text now truncates to the established MAX_STRING semantics for datatype strings instead of killing the client.

### DIFF
--- a/src/main/datatypes/MQ2WindowType.cpp
+++ b/src/main/datatypes/MQ2WindowType.cpp
@@ -504,16 +504,16 @@ bool MQ2WindowType::GetMember(MQVarPtr VarPtr, const char* Member, char* Index, 
 		if (pWnd->GetType() == UI_STMLBox)
 		{
 			CStmlWnd* cstmlwnd = static_cast<CStmlWnd*>(pWnd);
-			strcpy_s(DataTypeTemp, cstmlwnd->GetSTMLText().c_str());
+			strncpy_s(DataTypeTemp, cstmlwnd->GetSTMLText().c_str(), _TRUNCATE);
 		}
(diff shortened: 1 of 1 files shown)


Knightly1:
The resolution of this is not truncation, it's allowing the Window Text return to exceed 2048 characters. Otherwise the window text isn't accurate. Fixing this would be more about enabling that behavior without creating a larger fixed buffer.

PR #1023, opened 2026-07-05
Compute Me.NumGems from character data instead of the UI

### LINKED ISSUE #868: Me.NumGems is dependent on UI component
Implementation of CharacterMembers::NumGems in MQ2CharacterType.cpp, the implementation will count the number of visible gems.

The actual implementation should be to count the total applications of SPA_ADD_SPELL_SLOTS across AA, Gear and buffs.

CharacterZoneClient::CanUseMemorizedSpellSlot already does this, but it does the comparison, and we want the raw number of slots.

CodeCompiler:
Fixes #868

### Bug

`${Me.NumGems}` counted visible `CSPW_Spell#` children of the `CastSpellWnd`, so it returns 8 whenever the spell gem UI component isn't loaded/visible — exactly the dependency the issue calls out.

### Fix

Compute it from character data, as prescribed in the issue body: 8 base gems + the character's total `SPA_ADD_SPELL_SLOTS` (SPA 326) across items, AA, and buffs:

```cpp
Dest.DWord = std::min(8 + pLocalPC->TotalEffect(SPA_ADD_SPELL_SLOTS, true, 0, true, true), NUM_SPELL_GEMS);
```

This mirrors the accepted pattern `GetCharMaxBuffSlots()` already uses for `SPA_ADD_BUFF_SLOTS`. The `std::min` clamp to `NUM_SPELL_GEMS` preserves the old implementation's implicit upper bound so scripts iterating `Me.Gem[1..NumGems]` can't go out of range. `pLocalPC` is null-guarded at `GetMember` entry.

### DIFF
--- a/src/main/datatypes/MQ2CharacterType.cpp
+++ b/src/main/datatypes/MQ2CharacterType.cpp
@@ -3741,23 +3741,8 @@ bool MQ2CharacterType::GetMember(MQVarPtr VarPtr, const char* Member, char* Inde
 		return true;
 
 	case CharacterMembers::NumGems:
-		Dest.DWord = 8;
-		if (pCastSpellWnd)
-		{
-			char szWnd[32] = { 0 };
-			for (int i = 8; i < NUM_SPELL_GEMS; i++)
-			{
-				sprintf_s(szWnd, "CSPW_Spell%d", i);
-				if (CXWnd* wnd = pCastSpellWnd->GetChildItem(szWnd))
-				{
-					if (wnd->IsVisible() == 1)
-					{
-						Dest.DWord++;
-					}
-				}
-			}
-		}
-
+		// Number of available spell gems: 8 base plus any from SPA_ADD_SPELL_SLOTS (items, AA, buffs).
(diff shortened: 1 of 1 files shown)


dannuic:
I don't think this is an issue as implemented currently. The pattern of calculating instead of just reading directly from the UI is more fragile. The UI is guaranteed to always have the correct number of gems, the SPA calculation can change any day and break a calculation-based method.

CodeCompiler:
That's a fair point, and I've reworked it to a hybrid rather than replacing the UI read.

The window stays the authoritative source when it's loaded — the original `CSPW_Spell` count is unchanged in that path, so you keep the guaranteed-correct value and none of the SPA fragility in normal play. The calculation now only runs as a fallback when `pCastSpellWnd` isn't loaded, which is the specific case #868 reports (there the old code fell through to just the base 8). So the fragile path is only reached when there's no UI to read anyway — worst case it's wrong in a state where the previous behavior was already wrong.

If you'd still rather not carry the SPA fallback at all (i.e. you consider the "window not loaded" state a non-issue in practice), I'm happy to close this instead — your call.

brainiac:
pCastSpellWnd will always exist if you're in game. This change ends up doing nothing.

brainiac:
there isn't any need to make any changes here, so we can leave it as is

PR #1025, opened 2026-07-05
/itemnotify: select tribute donations instead of picking up the item

### LINKED ISSUE #874: [BUG] /itemnotify "ItemName" leftmouseup while tribute donation window open and bags closed will pickup item
**Describe the bug**
When using /itemnotify "ItemName" leftmouseup with bags closed and the tribute donation window open will pickup the item. With bags open it will properly select the item as a tribute donation.

**To Reproduce**
1. Go to Personal or Guild Tribute Master.
2. With bags closed issue the command /itemnotify "Some Item" leftmouseup
3. Item will be picked up onto cursor.

4. If bags are open the item will properly be selected as a tribute donation.

**Expected behavior**
Even with bags closed it should add the item to the Tribute Manager's donation window.

**MQ Version**
- [x] Live
- [x] Test
- [ ] Emu (ROF2)

CodeCompiler:
Fixes #874

### Bug

With the tribute donation window open and bags **closed**, `/itemnotify "some item" leftmouseup` picks the item up onto the cursor. With bags **open**, the same command correctly selects it as a tribute donation.

Cause: when the item's `CInvSlot` window doesn't exist (bag closed), `ItemNotify()` falls back to `PickupItem()`, which bypasses the client's slot-click handling — it special-cases the merchant window but has no tribute case. With bags open the command instead reaches `SendWndClick2` on the real slot window, and the client itself performs the donation selection.

### Fix

In both `PickupItem` fallback paths (named item, and `/itemnotify in <bag> <slot>`): when `pTributeMasterWnd` is visible, the cursor is empty, and the item is in possessions — open the item's container and send the click to the actual slot window, the same open-container + `SendWndClick2` technique the right-click spell-memming fallback in this function already uses. The container is left open so the selection remains valid until the user clicks Donate. If the slot still can't be resolved, the command reports an error instead of silently doing the wrong thing.

Behavior with an item already on the cursor is unchanged (drop). Note per eqlib's window table the client uses the same `TributeMasterWnd` screen for personal and guild tribute, so one check covers both.

### DIFF
--- a/src/main/MQ2Windows.cpp
+++ b/src/main/MQ2Windows.cpp
@@ -1467,6 +1467,25 @@ void ItemNotify(PSPAWNINFO pChar, char* szLine)
 					return;
 				}
 
+				// If the tribute master window is open, a left click should select the item as
+				// a donation instead of picking it up. That requires a real click on the slot's
+				// window, so open the container and send the click to it. The container is left
+				// open so that the selection remains valid until the donation is made.
+				if (pTributeMasterWnd && pTributeMasterWnd->IsVisible() && !ItemOnCursor()
+					&& globalIndex.GetLocation() == eItemContainerPossessions)
+				{
+					OpenContainer(pContainer, true);
+
+					CInvSlot* pInvSlot = GetInvSlot(globalIndex);
+
+					if (!pInvSlot || !pInvSlot->pInvSlotWnd || !SendWndClick2(pInvSlot->pInvSlotWnd, szNotification))
+					{
+						WriteChatf("Could not select item in %s %s for tribute donation.", szArg2, szArg3);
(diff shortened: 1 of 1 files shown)


brainiac:
Will need to think about this, as this is a change in behavior. I don't actually know with 100% certainty that this is how we want to "fix" it.

CodeCompiler:
Understood let me know what you would think would be the best solution for it as I am just back now a few days.

brainiac:
> Understood let me know what you would think would be the best solution for it as I am just back now a few days.

don't bend yourself out of shape over the timeliness of any of these, most of this stuff has been broken for over a decade, so theres no rush :)

PR #1026, opened 2026-07-05
Warn when MQIniPath points at a missing ini

### LINKED ISSUE #922: Notify user if custom locations in macroquest.ini is not found
if you have custom locations in your macroquest.ini like MQIniPath or ConfigPath and those locations no longer exist (copied your macroquest.ini from another computer, changed drives, etc) you will just silently fail - not even check default locations, just silently do nothing at all.
DebugConsole doesn't even give you *anything*
```ini
[MacroQuest]
MQIniPath=E:\OldComputer\MQ\MacroQuest.ini
ConfigPath=E:\OldComputer\MQ\config
```

I would expect that if you have custom locations listed, and they're not found, that debugconsole would tell you that the location wasn't found and then for it check default location and load those

CodeCompiler:
Fixes #922

### Problem

The directory keys (`ConfigPath`, `MacroPath`, …) already warn via MessageBox when their locations can't be found or created. But an `MQIniPath` redirect pointing at a **nonexistent** ini (the reporter's case: settings moved from an old computer) is silently accepted in `GetCreateMacroQuestIni` — every subsequent `GetPrivateProfileString` against it just returns defaults, with nothing in the console or a dialog to say why.

### Change

`GetCreateMacroQuestIni` is the single shared init point (used by both the loader's `InitializePaths` and the injected dll's `InitConfig`), so the check lives there: if a custom `MQIniPath` was set and the resolved ini doesn't exist, show the same style of MessageBox the directory-key failures already use:

> MQIniPath set in C:\MQ2\MacroQuest.ini could not be found: E:\OldPC\MQ\MacroQuest.ini

Warn-only by design — it deliberately does **not** fall back to the local ini, since pointing `MQIniPath` at a not-yet-created file and letting `WriteAllConfig` create it is a legitimate workflow, and a silent fallback would be its own surprise. If you'd prefer a fallback (or a DebugSpew instead of a dialog in the injected context), happy to adjust.

### DIFF
--- a/include/mq/base/Config.h
+++ b/include/mq/base/Config.h
@@ -363,8 +363,9 @@ inline std::string GetCreateMacroQuestIni(const std::filesystem::path& pathMQRoo
 			// If a custom MQIniPath was set but doesn't exist, warn the user since settings will not be loaded from it
 			if (pathMQini != pathFoundIni && !std::filesystem::exists(pathMQini, ec))
 			{
-				const std::string strTemp = "MQIniPath set in " + pathFoundIni.string() + " could not be found: " + pathMQini.string();
-				MessageBox(nullptr, strTemp.c_str(), "MacroQuest", MB_OK);
+				char szMessage[MAX_PATH * 2];
+				sprintf_s(szMessage, "MQIniPath set in %s could not be found: %s", pathFoundIni.string().c_str(), pathMQini.string().c_str());
+				MessageBox(nullptr, szMessage, "MacroQuest", MB_OK);
 			}
 		}
 
--- a/include/mq/base/Config.h
+++ b/include/mq/base/Config.h
@@ -343,6 +343,8 @@ inline std::string GetCreateMacroQuestIni(const std::filesystem::path& pathMQRoo
 
 		if (std::filesystem::exists(pathMQini, ec))
 		{
+			const std::filesystem::path pathFoundIni = pathMQini;
+
 			// Check to see if there is a different MacroQuest.ini we should be looking at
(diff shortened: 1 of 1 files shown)


include/mq/base/Config.h
```diff
@@ -357,6 +359,13 @@ inline std::string GetCreateMacroQuestIni(const std::filesystem::path& pathMQRoo
 			{
 				pathMQini = pathMQini / "MacroQuest.ini";
 			}
+
+			// If a custom MQIniPath was set but doesn't exist, warn the user since settings will not be loaded from it
+			if (pathMQini != pathFoundIni && !std::filesystem::exists(pathMQini, ec))
+			{
+				const std::string strTemp = "MQIniPath set in " + pathFoundIni.string() + " could not be found: " + pathMQini.string();
+				MessageBox(nullptr, strTemp.c_str(), "MacroQuest", MB_OK);
```
brainiac: this creates needless temporaries, we don't even store the string, so it would be simpler to just sprintf_s it to a local buffer and use that.

CodeCompiler:
Done - sprintf_s into a local buffer now, no std::string temporaries.

Knightly1:
I'm not sure I agree with this change. While the ini wouldn't be read immediately and defaults would be loaded, they're already sane defaults and any writes will create the file in place at the ini path specified. A message box is interruptive for an operation that essentially has no impact.

The issue this references is actually about not being able to create it in a location that could not be reached, which is a different problem than it not existing (even though that's what the issue says).

dannuic:
I'm with knightly here. There's no need to warn because this is an intended feature -- write the ini when it doesn't exist.

PR #1027, opened 2026-07-05
Strip stml anchor tags from event text

### LINKED ISSUE #682: Faction links don't get stripped from events
This still seems broken for faction links in faction text. I understand it has been mentioned and important parties are privy. But in the past it was also mentioned to open a ticket here. So here it is. :)

CodeCompiler:
Fixes #682

### Bug

Faction adjustment messages arrive as STML anchors with **no** `\x12` marker (formats captured by @brainiac in the issue):

```
Your faction standing with <a Faction="251">Frogloks of Guk</a> has been adjusted by -2.
```

Both event feeds gate link-stripping on `\x12` — `CheckChatForEvent` (`MQ2DataVars.cpp`) for macro `#event`s, and `LuaEventProcessor::Process` (`LuaEvent.cpp`) for `mq.event` — so the raw markup reaches the event engines and events written against the human-readable text never fire. `CleanItemTags` is a client function that only understands `\x12` tags, and `STMLToPlainText` is unsafe on stray `<`/`&`, so neither can be reused here.

### Fix

New `StripStmlAnchorTags()` in `MQ2Utilities.cpp` (exported like `StripMQChat`, since the lua plugin links against MQ2Main): removes `<a ...>text</a>` pairs in place, keeping the inner text. It only fires on well-formed pairs (malformed input breaks out untouched), removes the closing tag before the opening one so pointers stay valid, strictly shortens the string each iteration, and never overruns.

Applied in **both** event paths:
- `CheckChatForEvent` — after the existing `\x12` handling, so macro `#event`s see clean text.
- `LuaEvent.cpp` — anchor-bearing lines now take the has-links path, with the stripped feed (`m_blechStripped`) cleaned. Events registered with `keepLinks = true` intentionally keep the raw anchor, mirroring their existing `\x12` semantics, so Lua scripts can still parse the `Faction` id.

Chat display is unaffected — the chat window trampoline still receives the original message with the clickable link; only the event feeds change.

### DIFF
--- a/src/main/MQ2DataVars.cpp
+++ b/src/main/MQ2DataVars.cpp
@@ -878,6 +878,10 @@ void CheckChatForEvent(const char* szMsg)
 		strcpy_s(szClean, len + 64, out.c_str());
 	}
 
+	// Faction messages contain stml anchor tags (for example <a Faction="251">Frogloks of Guk</a>)
+	// instead of \x12 links. Strip the tags, keeping the inner text.
+	StripStmlAnchorTags(szClean);
+
 	strncpy_s(EventMsg, szClean, MAX_STRING - 1);
 	EventMsg[MAX_STRING - 1] = 0;
 	if (pMQ2Blech)
--- a/src/main/MQ2Main.h
+++ b/src/main/MQ2Main.h
@@ -198,6 +198,7 @@ char* GetEQPath(char* szBuffer, size_t len);
 MQLIB_API DWORD MQToSTML(const char* in, char* out, size_t maxlen = MAX_STRING, uint32_t ColorOverride = 0xFFFFFF);
 MQLIB_API void StripMQChat(const char* in, char* out);
 MQLIB_OBJECT void StripMQChat(std::string_view in, char* out);
+MQLIB_API void StripStmlAnchorTags(char* szText);
 MQLIB_API void STMLToPlainText(char* in, char* out);
 MQLIB_API char* GetSubFromLine(int Line, char* szSub, size_t Sublen);
 MQLIB_API const char* GetFilenameFromFullPath(const char* Filename);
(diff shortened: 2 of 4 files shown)


brainiac:
Two concerns:
- This is a hot code path, spending multiple passes to filter strings isn't very efficient.
- Theres a possibility that some people might be relying on these links since this has been this way for a very long time, so some research might be required into use cases and solution before this is merged.

CodeCompiler:
On the performance point — pushed a rewrite. The common case (no `<` in the line at all, i.e. virtually every chat line) now exits on a single `strchr`; the `"<a "` probe only runs when a `<` exists, and the strip itself is one forward copy pass with no `memmove` shuffling. So the added cost on the hot path is one `memchr`-class scan per fed line.

On compatibility, what I could establish:

- **Every `\x12` link type — including `ETAG_FACTION` when it arrives in `\x12` form — is already stripped to its inner text** by the existing `CleanItemTags` pass before events see it. The anchor-form faction message is the one format that leaks raw markup into event text, so this change aligns it with the semantics every other link already has. A script matching the raw `<a Faction=...>` markup is matching an accident of transport format that never renders in the chat window.
- **Lua has an explicit escape hatch that this PR preserves**: events registered with `keepLinks = true` still receive the raw anchor (`m_currentLine` is untouched), same as they do for `\x12` links — so Lua scripts that want the faction id keep a supported path.
- **The residual risk surface is macro `#event`s**, which have no `keepLinks` equivalent — a pattern deliberately written against the raw markup would stop matching. I couldn't find such a pattern in anything I could search, and the issue itself exists because patterns written the natural way (against the readable text) don't fire today.

If you'd like, I can split it — land the Lua half (escape hatch exists) and hold the macro half — or leave this open while you canvass use cases. Whatever's easiest for you.

PR #1029, opened 2026-07-05
Classify failing type methods as failures, not missing members

### LINKED ISSUE #329: No such 'bool' member 'Select'
In GitLab by @brainiac on Jun 21, 2021, 06:38

Had a wierd issue reported where this message `No such 'bool' member 'Select'` displayed in this line:

```
/delay 5s ${Window[TaskSelectWnd].Open} || ${Task[${TaskGetName}].Select}
```

image

Since .Select is a Task type, I don't think it makes sense to report that error. Select is a method, though. so there might be some wierdness related to that.

CodeCompiler:
Fixes #329

### Bug

Calling an existing type **method** that returns false — e.g. `${Window[TributeMasterWnd].Child[TMW_DonateList].Select[2]}` on a window type where `Select` bails (non-Listbox/TreeView/Combobox, or out-of-range index) — prints the misleading:

```
No such 'window' member 'Select'
```

`EvaluateMacroDataMember` classifies NotFound vs Failure using only `FindMember()`/`InheritedMember()`, but methods live in a separate `MethodMap` only reachable via `FindMethod()` — so an existing-but-failing method is reported as nonexistent. This is the surviving half of the issue (@brainiac's item 1, "Methods should not return false unless the method doesn't exist"); item 2 (the `'bool'` type-name inaccuracy) was already fixed by the `pType` capture refactor, and `MQ2TaskType::Select` got an interim always-true workaround.

### Fix

Include `type->FindMethod(Member)` in both existence checks (the `checkFirst` extension branch and the post-`GetMember` classification). Effects:

- A failing method now substitutes NULL silently (`EvaluateResult::Failure`), consistent with how failing members behave — no more bogus "No such member" error.
- Secondary fix for free: in the `checkFirst` branch (plugin type extensions), extension-registered **methods** were rejected as NotFound before the extension's `GetMember` was ever invoked — they were completely undispatchable. They now dispatch, matching the extensions-first design that extension members already have.

The exported `EvaluateMacroDataMember` now returns 0 (fail) instead of -1 (not found) for failing methods, which better matches its documented contract; the only in-repo consumer (`lua_MQMacroData.cpp`) checks `== 1` and is unaffected.

### DIFF
--- a/src/main/MQDataAPI.cpp
+++ b/src/main/MQDataAPI.cpp
@@ -392,7 +392,7 @@ MQDataAPI::EvaluateResult MQDataAPI::EvaluateMacroDataMember(MQ2Type* type, MQVa
 
 	if (checkFirst)
 	{
-		if (!type->FindMember(Member) && !type->InheritedMember(Member))
+		if (!type->FindMember(Member) && !type->FindMethod(Member) && !type->InheritedMember(Member))
 		{
 			return EvaluateResult::NotFound;
 		}
@@ -406,7 +406,7 @@ MQDataAPI::EvaluateResult MQDataAPI::EvaluateMacroDataMember(MQ2Type* type, MQVa
 		return EvaluateResult::Success;
 	}
 
-	if (!type->FindMember(Member) && !type->InheritedMember(Member))
+	if (!type->FindMember(Member) && !type->FindMethod(Member) && !type->InheritedMember(Member))
 	{
 		return EvaluateResult::NotFound;
 	}
--- a/src/main/MQDataAPI.cpp
+++ b/src/main/MQDataAPI.cpp
@@ -594,6 +594,9 @@ bool MQDataAPI::EvaluateDataExpression(MQTypeVar& Result, const char* pStart, ch
 		auto result = EvaluateMacroDataMember(pType, std::move(VarPtr), Result, pStart, pIndex, false);
 		if (result == EvaluateResult::NotFound)
 			MQ2DataError("No such '%s' member '%s'", pType->GetName(), pStart);
+		// a method that exists but failed reports accurately instead of as a missing member
+		else if (result == EvaluateResult::Failure && pType->FindMethod(pStart))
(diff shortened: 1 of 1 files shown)


Knightly1:
This is a pretty impactful and possibly breaking change. Someone will have to look at it further.

dannuic:
I'm not sure this does what it says it does, anyway. The "misleading" message here (which is correct IMO) is printed as a result of NotFound returning, all this PR does is short circuit to the NotFound result, which can only result in _more_ of that message, not _less_.

CodeCompiler:
Happy to walk through the flow I'm seeing here — correct me if I've misread it.

The "No such member" message only prints when `EvaluateMacroDataMember` returns `NotFound` (MQDataAPI.cpp:595). This PR adds `&& !FindMethod(Member)` to the two conditions that *return* `NotFound` (lines 395, 409), which makes that path harder to reach, not easier. Since `(A && B && C)` implies `(A && B)`, the inputs that hit `NotFound` after the change are a subset of before — so the message fires less often or the same, never more. For `${Window[x].Select[99]}`: `Select` is a method, so `GetMember` runs it and it returns false; at line 409 `!FindMember` and `!InheritedMember` are both true, so today it returns `NotFound` → the message, whereas with `!FindMethod("Select")` false it returns `Failure`.

On the good points, though — the impact concern is fair, this is on the hot path for every type. And @dannuic, you're right that just going silent loses signal. The real issue in #329 is that the message is *inaccurate* ("No such member 'Select'" when `Select` exists), not that a message exists — so I pushed a follow-up commit: a method that exists but fails now reports accurately as `'window' method 'Select' failed` on the `Failure` path, instead of the misleading wording or going silent. That keeps the diagnostic you want, corrected.

Two notes either way:
- A method defined only on a *parent* type still isn't covered, since `InheritedMember` checks the parent's members but not its methods.
- Side effect worth being aware of: methods registered on type extensions currently never dispatch at all (they short-circuit to `NotFound` before the method runs), and this makes them run.

Glad to take it whichever direction you'd prefer.

PR #1030, opened 2026-07-05
Make Lua actor message handling robust to yields and errors

### LINKED ISSUE #842: [BUG]
**Describe the bug**
If you have a bug in your actor message handler, you can get into this state:
``` 
Lua Actor Failure:
d:\mq\lua\aqo\utils\lootutils.lua:257: bad argument #1 to 'ipairs' (table expected, got nil)
```
initial error, then subsequent messages:
```
Lua Actor Failure:
cannot resume non-suspended coroutine
```

**To Reproduce**
1. Send a message to an actor with an empty table which gets serialized to nil
2. Try to iterate the table with ipairs in the message handler

**Expected behavior**
Subsequent messages shouldn't be blocked from processing? Not sure. In my case, it was an intermittent thing because one time the table content of my message had no items in it. Subsequent messages would have worked properly.

**MQ Version**
- any

### LINKED ISSUE #854: [BUG] Possible Event Overrun?
**Describe the bug**
When sending messages to a Actor that handles messages by replying to the sending Actor, after a few minutes, the receiving client will crash.  This was test code that I and AQ1 both were running separately, yet within minutes of each other had the same result.

**To Reproduce**
1. On the receiving client, run the following script.
```lua
mq = require('mq')
local actors = require('actors')

local function message_handler(message)
   local func = loadstring("return " .. message.content)
   if func then
      local response = func()
      message:send(response)
   else
      printf("Message failed: %s", message.content)
   end
end

actors.register(message_handler)

while true do
   mq.delay(1000)
end
```
2. On the sending character run this script:  (note the character name will need to be adjusted this is for example only)
```lua
local mq = require('mq')
local actors = require('actors')

local a = actors.register(function(message) print(message.content) end)

while true do
   mq.delay(1000)
   a:send({character="bubba", script='cb1'}, "mq.TLO.Me.Name()")
end
```
3.  Wait a few minutes (1-5 minutes)
4.  The Receiving client will crash (with popup).  Brainiac provided a dump that I've attached below.

**Expected behavior**
No crash.

**Screenshots**

**MQ Version**
- [X] Live
- [X] Test
- [ ] Emu (ROF2)

**Additional context**
[dump.txt](https://github.com/macroquest/macroquest/files/14818156/dump.txt)

CodeCompiler:
Fixes #842
Fixes #854

Two defects in the same few lines of `LuaDropbox::Receive` — which resumes one persistent coroutine for every incoming actor message — with compounding symptoms. One PR because the edits interlock in the same function.

### #854 — VM corruption crash (dangling `lua_State` in `ParseHeader`)

The forced-yield debug hook (`LuaThread::YieldAt` → `lua_forceYield`) is left armed when the post office delivers a message. Under LuaJIT, debug hooks live in the `global_State` — they fire in **every** coroutine of the VM, not just the script coroutine they were set on — and `lua_forceYield` never consults `GetAllowYield()`, so the `ScopedYieldDisabler` can't stop it. When the leftover instruction budget expires mid-handler, the handler coroutine yields silently (sol treats `call_status::yielded` as `valid()`), and the **next** message pushes a fresh function + arguments onto the *suspended* coroutine and resumes into the middle of the interrupted handler — corrupting the stack, with the crash surfacing later exactly as in the issue's dump (`Receive` → `ParseHeader` → `state_view(nullptr)` → AV). The probabilistic 1–5 minute window at ~1 msg/sec matches the hook-expiry timing.

The two sibling paths that resume side coroutines already unset the hook first, with a comment saying why (`LuaDropbox::Process`, `LuaImGuiProcessor::Pulse`). `Receive` was the one resume path missing the guard — a regression from 7820f506 ("Localized the hook unset to actors"), which removed the hook-clearing that `SetAllowYield(false)` used to do and re-added it only to `Process`. This restores the lost protection in the established pattern; the hook is re-armed by `YieldAt(m_turboNum)` at the next script resume.

### #842 — one handler error permanently breaks the actor

When the handler raises a Lua error, the coroutine's thread enters an error status that LuaJIT permanently refuses to resume — every subsequent message fails with `cannot resume non-suspended coroutine`. After a failed invocation, recreate `m_thread`/`m_coroutine` exactly as the constructor builds them, so the next message gets a fresh resumable thread. `ScopedYieldDisabler` holds its own `shared_ptr` and `result.abandon()` runs before the reassignment, so nothing touches the discarded thread. (The issue's side note about empty tables serializing to nil is a separate `SerializeProto` design question, intentionally not touched.)

### DIFF
--- a/src/plugins/lua/LuaActor.cpp
+++ b/src/plugins/lua/LuaActor.cpp
@@ -457,18 +457,35 @@ void LuaDropbox::Receive(const std::shared_ptr<Message>& message)
 {
 	try
 	{
-		ScopedYieldDisabler disableYield(LuaThread::get_from(m_thread.state()));
+		std::shared_ptr<LuaThread> luaThread = LuaThread::get_from(m_thread.state());
+
+		// remove any existing hooks, they will be re-installed when running in onpulse.
+		// this prevents the forced-yield hook from suspending the message callback mid-
+		// execution, which would leave the coroutine suspended and corrupt it when the
+		// next message resumes it with a freshly pushed function and arguments.
+		if (luaThread)
+			lua_sethook(luaThread->GetLuaThread().lua_state(), nullptr, 0, 0);
+
+		ScopedYieldDisabler disableYield(luaThread);
 
 		sol::function_result result = m_coroutine(LuaMessage(this, message));
 		if (!result.valid())
 		{
 			LuaError("Lua Actor Failure:\n%s", sol::stack::get<std::string>(result.lua_state(), result.stack_index()).c_str());
 			result.abandon();
+
+			// an error in the callback leaves the coroutine's thread in a state that can never
+			// be resumed again, so recreate it to allow subsequent messages to be processed
+			m_thread = sol::thread::create(m_parentThread.state());
+			m_coroutine = sol::coroutine(m_thread.state(), m_callback);
 		}
 	}
 	catch (std::runtime_error& e)
 	{
 		LuaError("Lua Actor Failure:\n%s", e.what());
+
+		m_thread = sol::thread::create(m_parentThread.state());
+		m_coroutine = sol::coroutine(m_thread.state(), m_callback);
 	}
 }
 


dannuic:
I think the first change is correct, but I have major concerns about re-creating lua threads after they are shut down. It seems like a pretty big anti-pattern and the correct solve for that is likely more along the lines of handling error cases when they happen.

This change will also need a lot of testing to ensure that it doesn't have secondary and tertiary effects.

PR #1031, opened 2026-07-05
Clear the switch target when it leaves the switch manager

### LINKED ISSUE #857: [BUG] Switch target may remain one additional frame after being cleared
Switches don't normally get deallocated, except on zoning.

There is a race where rarely the switches can be deallocated before the final pulse before zoning. Maybe we should check where they get unloaded and ensure that the switch target is cleared at that point.

CodeCompiler:
Fixes #857

### Bug

Switches are deallocated when the old zone unloads, and (as the issue notes) that can happen **before** MQ detects it is zoning. The only automatic clear of `pSwitchTarget` is the `SetSwitchTarget(nullptr)` in the `WereWeZoning` reset block — i.e. after zoning *completes* — so for the intervening pulse(s), `${Switch...}`, `/face door`, plugins reading `pSwitchTarget`, and the switch inspector can dereference a freed `EQSwitch`.

Ground items already have proactive invalidation via the `EQGroundItemListManager` Clear/Delete detours; switches have no equivalent teardown hook, and adding one would require a new eqlib offset.

### Fix

Offset-free: validate by **pointer identity** each heartbeat. If `pSwitchTarget` is no longer present in `pSwitchMgr->Switches[]`, clear it via `SetSwitchTarget(nullptr)` (which also clears the deprecated `pDoorTarget`). The stale pointer is never dereferenced — only compared. Every legitimate switch target originates from the manager (`GetSwitchByID`/`FindSwitchByName`/the switch type's `Target` method all iterate it), so membership is a sound validity test; the scan is at most `NumEntries` pointer compares per heartbeat.

Placed before `DrawHUD()` so plugin `OnDrawHUD` callbacks are covered along with `Pulse`/`PulsePlugins`/macros/ImGui. Residual window that polling can't close: the render path (`ImGuiManager_DrawFrame`, `DrawNetStatus`) can still observe the pointer between the client's teardown and the next heartbeat — fully closing that would need the teardown-hook approach with a new eqlib offset; this covers the pulse-level race the issue reports without one.

### DIFF
--- a/src/main/MQ2Pulse.cpp
+++ b/src/main/MQ2Pulse.cpp
@@ -701,6 +701,31 @@ static HeartbeatState Heartbeat()
 
 	UpdateMQ2SpawnSort();
 
+	// Switches are deallocated when the zone unloads, which can happen before we
+	// detect that we are zoning. If the switch target is no longer in the switch
+	// manager, clear it so that we never hold a dangling pointer. See #857.
+	if (pSwitchTarget)
+	{
+		bool switchExists = false;
+
+		if (pSwitchMgr)
+		{
+			for (int i = 0; i < pSwitchMgr->NumEntries; ++i)
+			{
+				if (pSwitchMgr->Switches[i] == pSwitchTarget)
+				{
+					switchExists = true;
+					break;
+				}
+			}
+		}
+
+		if (!switchExists)
+		{
+			SetSwitchTarget(nullptr);
+		}
+	}
+
 	DebugTry(DrawHUD());
 	DebugTry(PulseMQ2AutoInventory());
 


brainiac:
Checking if doors changed every frame isn't the right fix for this. My original comment states that this should occur at the time the door is removed. That implies creating a new hook into the client at the right spot.

Working around that requirement isn't sufficient

CodeCompiler:
Fair enough â€” closing this. You're right that per-frame validation is a workaround for what should be an event: the correct shape is a detour at the point the client tears the switches down (the `EQGroundItemListManager::Clear`/`Delete` pattern), which needs a new eqlib offset for the teardown site rather than anything this repo alone can provide.

If it's useful when someone picks this up: from analysis of the current Live client we can supply the candidate teardown function (address + byte signature) for the eqlib side â€” happy to post that on #857.

PR #1034, opened 2026-07-21
Add custom buttons to Item Display

Hazado:
Replaces Lucy Button with custom URL button that defaults to Lucy 
Adds additional URL button that shows above the Lucy Button (Hidden if no URL is setup)

Add 6 custom buttons that sit next to modified button that can be used as custom action buttons
Each button can have a custom name, action (/mac Keepthisitem) and has a checkbox that will pickup the item before doing the action.
Custom buttons are hidden if you don't have any action/URL assigned to it
6 action button also will be hidden if the pickup item option is used and you don't have the item in your inventory.

Added three variables that will get replaced when the action/URL is run
%name% - Items name
%id% - Items ID
%count% - total count you have

New Ini entries with defaults shown
> Button01URL=https://lucy.allakhazam.com/item.html?id=%id%
> Button01Name=Lucy
> Button02URL=
> Button02Name=
> Cust01Name=Keep
> Cust02Name=Sell
> Cust03Name=Bank
> Cust04Name=Tribute
> Cust05Name=Ignore
> Cust06Name=Announce
> Cust01Action=
> Cust02Action=
> Cust03Action=
> Cust04Action=
> Cust05Action=
> Cust06Action=
> Cust01PickupItemEnabled=0
> Cust02PickupItemEnabled=0
> Cust03PickupItemEnabled=0
> Cust05PickupItemEnabled=0
> Cust04PickupItemEnabled=0
> Cust06PickupItemEnabled=0

Pictures
Item in my inventory
image

Item not in my inventory (Buttons 3 and 4 pickup option unchecked)
image

MQ Settings window
image

### DIFF
--- a/src/plugins/itemdisplay/MQ2ItemDisplay.cpp
+++ b/src/plugins/itemdisplay/MQ2ItemDisplay.cpp
@@ -24,6 +24,9 @@
 #include <string_view>
 #include <fstream>
 
+#include <regex>
+#include <misc/cpp/imgui_stdlib.h>
+
 using namespace mq::datatypes;
 
 PreSetup("MQ2ItemDisplay");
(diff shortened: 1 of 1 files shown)


brainiac:
This is a great idea but there's a pretty extreme level of copy paste going on here. This kind of pattern makes this extremely difficult to maintain because every change needs to be done 6 times. Adding a 7th would just make it worse. 

Typically in scenarios such as this we would use a container such as a vector and use an index to select which button to work with instead of copy pasting the same code 6 times. 

I would like to see this change get made -- rewrite this to parameterize the button index and eliminate the copy paste and we can see about next steps in merging this change.

Hazado:
> This is a great idea but there's a pretty extreme level of copy paste going on here. This kind of pattern makes this extremely difficult to maintain because every change needs to be done 6 times. Adding a 7th would just make it worse.
> 
> Typically in scenarios such as this we would use a container such as a vector and use an index to select which button to work with instead of copy pasting the same code 6 times.
> 
> I would like to see this change get made -- rewrite this to parameterize the button index and eliminate the copy paste and we can see about next steps in merging this change.

Right, my c++ abilities are amateur level
I will attempt this but it probably wont be what your looking for
I dont mind getting help on this

Hazado:
Alright, Github Copilot was used to refactor
Found or caused a lot of bugs when doing this related to the IMGUI and /itemdisplay commands.
I've tested it pretty extensively, but please review it with that in mind.

Testing done was changing all the imgui settings/itemdisplay commands and making sure the INI and item display window reflected it realtime.
Unloading/Loading the plugin with INI changes made.
Tested that holding the item or another item wont cause issues when it picks up items (wouldnt expect it because its using /itemnotify)

Changed the item pickup to check if the item is on the cursor and pickup the entire stack for items that are stacked

Hazado:
Removed some of the redundant code
There is probably more that can be done

Hazado:
I think this is ready for a review now

I might have went a bit overboard with the buttons :)
image
image

PR #1040, opened 2026-09-20
fix assignment for previous pr

Sicprofundus:
- also fixed incorrect assignment for Dar


src/main/datatypes/MQ2CharacterType.cpp
```diff
@@ -1350,7 +1350,7 @@ bool MQ2CharacterType::GetMember(MQVarPtr VarPtr, const char* Member, char* Inde
 		return true;
 
 	case CharacterMembers::PlatinumShared:
-		Dest.DWord = pLocalPC->BankSharedPlat;
+		Dest.Int64 = pLocalPC->BankSharedPlat;
```
dannuic: Use Dest.Set()

brainiac: Don't. Emu won't return int64 and that would break it

dannuic: Dest.Set<int64_t>(), these setters are deprecated

brainiac: Yeah if you don't infer the type that works

dannuic: reference: line 248 of MacroDataTypes.h

Sicprofundus:
It is really hard not to "well i touched this line, the line next to it needs updating" and then "wow, now i have to account for all mass in the universe, that is kinda daunting".

saw a SPELL* I wanted to change to EQ_Spell*, heh.

dannuic:
In this case, very specifically, your previous change didn't actually do anything (it was my fault for not noticing) because it was still only setting DWord, which is only 32-bits. You didn't actually fix the narrowing conversion, you only declared that you fixed it by changing the type stored in the variant -- so you also needed to fix the real conversion (the place where it set the variant to the value)

Sicprofundus:
> In this case, very specifically, your previous change didn't actually do anything (it was my fault for not noticing) because it was still only setting DWord, which is only 32-bits. You didn't actually fix the narrowing conversion, you only declared that you fixed it by changing the type stored in the variant -- so you also needed to fix the real conversion (the place where it set the variant to the value)

yeah brain pointed that out to me - the .set updates now fix that, correct? or i need to cast inside the set?

